PR #4, opened 2021-08-16
Wf 336 create spa

georgejolema:


src/app/app.module.ts
```diff
@@ -25,10 +28,21 @@ import { HttpClientModule } from '@angular/common/http';
...
     EntitlementsModule,
     DropdownMenuModule,
-    AvatarModule,
     IconModule,
+    LayoutModule,
+    LogoModule,
+    NgbDropdownModule,
+    AvatarModule,
+    WebSdkModule.forRoot({
+      cx: { // TODO: this is a hack to get entitlements work using websdk
```
alyahmedaly, line 37: Can you create a ticket to investigate how we can avoid this

src/app/auth.guard.ts
```diff
@@ -5,13 +5,23 @@ import { OAuthService } from 'angular-oauth2-oidc';
...
+  ) {}
 
   canActivate(
-    next: ActivatedRouteSnapshot,
-    state: RouterStateSnapshot): Observable<boolean | UrlTree> | Promise<boolean | UrlTree> | boolean | UrlTree {
+    _next: ActivatedRouteSnapshot,
+    _state: RouterStateSnapshot
+  ): boolean {
+    // TODO: Uncomment out when login is ready see ticket WF-343
+    // if (this.oauthService.hasValidAccessToken()) {
```
alyahmedaly, line 19: I would keep this link in place, no need to comment it out

src/assets/entitlements.json
```diff
@@ -0,0 +1,744 @@
+[
```
alyahmedaly, line 1: I wonder if we need the whole file.

src/app/app.component.html
```diff
@@ -1,293 +1,104 @@
...
+<nav class="bb-layout__topbar">
+  <!-- Skip to content -->
+  <button
+    i18n-aria-label="
+      Move focus to beginning of content|Value for aria-label saying that this button will move the focus to the
+      beginning of the content
+    "
+    aria-label="skip to content"
+    class="bb-layout__skip-to-content"
+    (click)="$any($event).view?.window?.document?.querySelector('[role=\'main\']')?.focus()"
```
alyahmedaly, line 17: Can you move this logic to ts file, this looks hacky to me

projects/transfer-journey/src/lib/transfer-journey.module.ts
```diff
@@ -13,6 +13,7 @@ import { TransferJourneyComponent } from './transfer-journey.component';
 import { MakeTransferSuccessViewComponent } from './views/make-transfer-success-view.component';
 import { MakeTransferSummaryViewComponent } from './views/make-transfer-summary-view.component';
 import { MakeTransferViewComponent } from './views/make-transfer-view.component';
+import { EntitlementsGuard, EntitlementsModule } from '@backbase/foundation-ang/entitlements';
```
yenobi, line 16: Import without actually using in the file?

src/app/login/login-page/login-page.component.ts
```diff
@@ -3,16 +3,13 @@ import { OAuthService } from 'angular-oauth2-oidc';
 
 @Component({
   selector: 'app-login-page',
-  templateUrl: './login-page.component.html',
-  styleUrls: ['./login-page.component.sass']
```
yenobi, line 7: Looks like the file itself was not removed.

src/app/auth.guard.ts
georgejolema: that code does not work. After a while the app gets stuck and It is no longer possible to login. It needs more investigation

src/app/app.module.ts
georgejolema: I know the reason. WebSDK is not supposed to handle entitlements like that. Those configurations are given by lean-client. But the hack was possible because WebSDK can send a whole configuration object to lean-client. The question here is should we add that functionality to foundation-ang?

src/assets/entitlements.json
georgejolema: not really, I can remove the other things

src/app/app.component.html
georgejolema: I am following the best practices provided by Web-UI team. This is the community documentation that I followed https://designsystem.backbase.com/v2/web-components/horizontal and this is supposed to be the Backbase way. Of course we can make it better and contribute.

src/assets/entitlements.json
georgejolema: done

src/app/app.component.html
georgejolema: fixed

projects/transfer-journey/src/lib/transfer-journey.module.ts
georgejolema: fixed

src/app/login/login-page/login-page.component.ts
georgejolema: fixed

santam85 (approved):
- [x] Let's fix the auth problem before merging

yenobi (approved):

PR #5, opened 2021-08-25
Angular update

santam85:
Cleaned up a bit of unused code and updated dependencies


projects/transactions-journey/src/lib/transactions-journey-config.service.ts
```diff
@@ -1,28 +1,6 @@
...
-  }
-
-  get defaults(): TransactionsJourneyConfiguration {
-    return configDefaults;
-  }
-
-  get pageSize(): number {
-    return this.config.pageSize;
-  }
+export class TransactionsJourneyConfiguration {
```
kirillsud, line 4: 👍  I like this refactoring. What was the reason of refactoring? Was the previous implementation not working with Angular 12?

kirillsud (approved):

projects/transactions-journey/src/lib/transactions-journey-config.service.ts
santam85: I just wanted to make the code more concise! It was working ok even without it.
Journey authors liked the simplification too.

alyahmedaly (approved):

PR #9, opened 2021-09-05
Icons replacement examples

okadurin:
This is a solution for https://backbase.atlassian.net/browse/PXS-3683


georgejolema (approved):

src/app/app.component.html
```diff
@@ -59,6 +59,7 @@
           [attr.aria-current]="myAccountsRouterLink.isActive ? 'page' : undefined"
           role="link"
         >
+          <bb-icon-ui class="bb-layout__horizontal-nav-item-icon" name="transactions" size="md"></bb-icon-ui>
```
mistermark, line 62: I would recommend to include more icons than just one in the app. Find places to put icons so the impact of changing them is more prominent

src/styles_replace-entire-icon-font-when-names-dont-match.scss
```diff
@@ -0,0 +1,22 @@
+@import "./styles";
+@import "~@backbase/backbase-theme/scss/main";
+
+// If the new font's symbol names don't match the names from "Material Icon" font, then we need 
+// to change the font-family for .bb-icon and "content" property for every icon to map the names
+
+// IconMoon for all icons
+// free version https://icomoon.io/icons-icomoon.html
+@font-face {
+  font-family: "Icon Moon";
```
mistermark, line 10: Change font-family name to "IcoMoon"

src/styles_replace-entire-icon-font-when-names-dont-match.scss
```diff
@@ -0,0 +1,22 @@
...
+  font-style: normal;
+  font-weight: 400;
+  src: url("assets/custom-icons/IcoMoon-Free.ttf") format("truetype");
+}
+
+.bb-icon {
+  font-family: "Icon Moon";
```
mistermark, line 17: Change font-family name to "IcoMoon"

src/styles_replace-icon-with-svg.scss
```diff
@@ -0,0 +1,27 @@
+@import "./styles";
+@import "~@backbase/backbase-theme/scss/main";
+
+// Overriding the icon
```
mistermark, line 4: I would prefer a more sustainable solution where it's easier for the project to manage their SVG icons. I see problems with having to replicate the same icon SVG file multiple times to get a range of colors.

Let's discuss a better solution with the team.

src/styles_replace-some-icons.scss
```diff
@@ -0,0 +1,22 @@
...
+@import "./styles";
+@import "~@backbase/backbase-theme/scss/main";
+
+// If the new font's symbol names don't match the names from "Material Icon" font, then we need 
+// to change the "font-family" and the "content" property for the icon we want to update
+
+// IconMoon for all icons
+// free version https://icomoon.io/icons-icomoon.html
+@font-face {
+  font-family: "Icon Moon";
```
mistermark, line 11: Change font-family name to "IcoMoon"

src/styles_replace-some-icons.scss
```diff
@@ -0,0 +1,22 @@
...
+  font-style: normal;
+  font-weight: 400;
+  src: url("assets/custom-icons/IcoMoon-Free.ttf") format("truetype");
+}
+
+.bb-icon-transactions {
+  font-family: 'Icon Moon';
```
mistermark, line 18: Change font-family name to "IcoMoon"

mistermark (commented):
Let's organize to discuss an alternative solution to the SVG icon situation.

src/app/app.component.html
okadurin: The project is in the early stage and will be filled out with more data and more components with icons. When it's stable we will update the project so more icons are replaced

mistermark (approved):

santam85:
Hey Oleksii, this goes a bit beyond the point of the golden sample repo. I am happy showing something like this in the DSO docs, but it's overkill for this repo. We can leave here ONE example of icon replacement, but not behind configuration or flags, just as an hardcoded override.

santam85 (changes requested):
Hey Oleksii, this goes a bit beyond the point of the golden sample repo. I am happy showing something like this in the DSO docs, but it's overkill for this repo. We can leave here ONE example of icon replacement, but not behind configuration or flags, just as an hardcoded override.

okadurin:
Close this PR based on Marco's feedback. Will think about alternative solution.

PR #10, opened 2021-09-07
Wf 337 Add entitlements to Transfer journey

georgejolema:


src/app/services/app-auth.service.ts
```diff
@@ -0,0 +1,158 @@
...
+    this.oauthService.events.subscribe((event) => {
+      if (event instanceof OAuthErrorEvent) {
+        console.error('OAuthErrorEvent Object:', event);
+      } else {
+        console.warn('OAuthEvent Object:', event);
+      }
+    });
+
+    /**
+     * After successful auh and the `token_received` event
```
kirillsud, line 36: auh => auth

src/app/services/app-auth.service.ts
```diff
@@ -0,0 +1,158 @@
...
+ * This service is build with external OIDC compliant OAuthService
+ * for demonstration purposes, be careful and read all of the comments
+ *
+ */
+@Injectable({
+  providedIn: 'root',
+})
+export class AppAuthService {
+  private isAuthenticated$$ = new BehaviorSubject<boolean>(false);
+  public isAuthenticated$ = this.isAuthenticated$$.asObservable();
```
kirillsud, line 17:
```suggestion
  public isAuthenticated$ = this.oauthService.events.pipe(map(() => this.oauthService.hasValidAccessToken()));
```

kirillsud (approved):

yenobi:
Did you merge master to your branch?

src/app/services/app-auth.service.ts
yenobi: As for me, it looks like an improvement and can be applied.
Please, also move the comment from the constructor, this one:
`After successful auth and the `token_received` event it will update isAuthenticated$ to activate the routes`

georgejolema: It does not let me apply it from here but I will add the change manually

src/app/services/app-auth.service.ts
georgejolema: fixed

src/app/services/app-auth.service.ts
georgejolema: fixed

src/app/services/app-auth.service.ts
```diff
@@ -14,7 +15,7 @@ import { BehaviorSubject } from 'rxjs';
 })
 export class AppAuthService {
   private isAuthenticated$$ = new BehaviorSubject<boolean>(false);
```
kirillsud, line 17: I think it could be removed now

georgejolema: it is still used by isDoneLoading$$

src/app/services/app-auth.service.ts
```diff
@@ -13,8 +14,13 @@ import { BehaviorSubject } from 'rxjs';
   providedIn: 'root',
 })
 export class AppAuthService {
-  private isAuthenticated$$ = new BehaviorSubject<boolean>(false);
-  public isAuthenticated$ = this.isAuthenticated$$.asObservable();
+  public isAuthenticated$ = this.oauthService.events.pipe(
+    map(() => this.oauthService.hasValidAccessToken()),
+    shareReplay({
```
georgejolema, line 19: @kirillsud and @sergii-backbase I was forced to add shareReplay to make it stay alive. Let me know if you agree

src/app/services/app-auth.service.ts
kirillsud: Where? I see only one place here 
image

src/app/app.component.ts
```diff
@@ -16,14 +19,12 @@ export class AppComponent {
...
 
-  public isAuthenticated$ = this.authService.isAuthenticated$;
-
   logout(): void {
     this.authService.logout();
   }
 
   focusMainContainer(event: MouseEvent) {
-    const element = event.view?.window?.document?.querySelector("[role='main']") as HTMLElement | undefined;
+    const element = event.view?.window?.document?.querySelector('[role=\'main\']') as HTMLElement | undefined;
```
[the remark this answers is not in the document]
georgejolema: we don't have prettier. It was just added by mistake. But the standard should be single quotes (just like backbase standard)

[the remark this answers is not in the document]
yenobi: Can you, please, point to me, where it is defined, I did not know about it?

yenobi (approved):

yenobi:
I have approved, but I think we should wait a little bit, before we merge this PR to provide the actual users from the identity service. I have already asked for help with figuring out how to do this in the support channel, then we can apply the different entitlements.

PR #11, opened 2021-09-10
Feature/wf 343  add authentication example

yenobi:


src/app/services/app-auth.service.ts
```diff
@@ -0,0 +1,70 @@
...
+   */
+  constructor(private http: HttpClient, private router: Router) {}
+
+  /**
+   * In the production version only `redirectUri` is needed to be provided as an argument for
+   * this.authServiceFromFoundation.login({ redirectUri }) method call
+   */
+  login({ login, redirectUri }: { login: string; redirectUri: string }): void {
+    this.http
+      .post('/api/auth/login', { login, redirectUri })
```
yash-backbase, line 44: Thank you @yenobi for this example. It was long pending.

One suggestion here perhaps - the auth service is only going to be used for hybrid strategy i.e delegating auth responsibility to Identity(or any other IDP). But in the example here you the Angular app is directly interacting with an authentication endpoint(giving it a feeling of bas auth strategy).

Is it worth mentioning here explicitly in the comment that in production this login mechanism will be handled by customer's IDP?

yenobi: I have updated the comment here https://github.com/Backbase/golden-sample-app/pull/11/commits/edcff60a036679fe2324b632d51ebb656d0eca33

Do you think it will be enough or should I have a bigger explanation?

yash-backbase: I think it is enough - thanks

src/app/login/login-page/login-page.component.html
```diff
@@ -1,2 +1,26 @@
...
+			[showErrors]="loginForm.controls.login.invalid && loginForm.controls.login.touched"
+			data-role="login-validation-message"
+		>
+            <span i18n="login is required and minimu leanth is 4 characters validation error">
+                Required field and should be at least 4 characters
+            </span>
+        </bb-input-validation-message-ui>
+		<bb-input-text-ui
+			inputClassName="mb-2"
+    		label="Username or email"
```
georgejolema, line 14: we need to add translation to the label

src/app/login/login-page/login-page.component.html
```diff
@@ -1,2 +1,26 @@
...
+    		formControlName="login"
+    		[required]="true"
+			data-role="login-input"
+  		></bb-input-text-ui>
+		<bb-input-password-ui
+			inputClassName="mb-2"
+      		label="Password"
```
georgejolema, line 21: we need to add translation to the label

src/app/login/login-page/login-page.component.ts
```diff
@@ -1,16 +1,29 @@
...
-  ) { }
+export class LoginPageComponent implements OnInit {
+  loginForm!: FormGroup;
 
-  login(): void {
-    this.oAuthService.initLoginFlow();
+  constructor(private authService: AppAuthService, private formBuilder: FormBuilder) {}
+
+  ngOnInit() {
+    this.loginForm = this.formBuilder.group({
```
georgejolema, line 15: this can be defined in the constructor or where the property is defined because you don't need to wait for the component to setup these values. By doing this, you will no longer need **ngOnInit**

src/app/services/app-auth.service.ts
```diff
@@ -0,0 +1,71 @@
...
+   */
+  constructor(private http: HttpClient, private router: Router) {}
+
+  /**
+   * In the production version only `redirectUri` is needed to be provided as an argument for
+   * this.authServiceFromFoundation.login({ redirectUri }) method call,
+   * and the authentication request will be handled by identity provided from the settings
+   */
+  login({ login, redirectUri }: { login: string; redirectUri: string }): void {
+    this.http
```
georgejolema, line 44: you should add an error handler for 401 response. Also I don't think you should just redirect from here. But also return something to confirm that login was successful or not.

src/app/login/login-page/login-page.component.html
```diff
@@ -1,2 +1,26 @@
-<p>login-page works!</p>
-<button (click)="login()">login</button>
+<form [formGroup]="loginForm" (ngSubmit)="onSubmit()" novalidate data-role="login-form">
+	<div class="col-lg-6">
+		<bb-header-ui headingType="h1" heading="Welcome to golden sample application"></bb-header-ui>
+		<bb-input-validation-message-ui
+			[showErrors]="loginForm.controls.login.invalid && loginForm.controls.login.touched"
+			data-role="login-validation-message"
+		>
+            <span i18n="login is required and minimu leanth is 4 characters validation error">
```
georgejolema, line 8: ui standards that I have seen in capabilities, usually put the error message below the input

src/app/login/login-page/login-page.component.html
```diff
@@ -1,2 +1,26 @@
...
+    		[required]="true"
+			data-role="login-input"
+  		></bb-input-text-ui>
+		<bb-input-password-ui
+			inputClassName="mb-2"
+      		label="Password"
+    	>
+    	</bb-input-password-ui>
+		<button class="mt-2" bbButton type="submit" aria-label="submit">submit</button>
+	</div>
```
georgejolema, line 25: there should be an error message if the login is not successful

src/app/services/app-auth.service.ts
```diff
@@ -0,0 +1,71 @@
...
+ * This is not the actual implementation of authentication service,
+ * this is only example, be careful with reading all of the notes inside.
+ */
+@Injectable({
+  providedIn: 'root',
+})
+export class AppAuthService {
+  private isAuthenticated$$ = new BehaviorSubject<boolean>(false);
+
+  private currentUser$$ = new BehaviorSubject<User | null>(null);
```
georgejolema, line 23: if I refresh the page, I lose my login. We should store that in localStorage/sessionStorage to keep the session alive.

src/app/guards/auth.guard.ts
```diff
@@ -0,0 +1,20 @@
...
+import { ActivatedRouteSnapshot, CanActivate, Router, RouterStateSnapshot } from '@angular/router';
+import { Observable } from 'rxjs';
+import { tap } from 'rxjs/operators';
+import { AppAuthService } from '../services/app-auth.service';
+
+@Injectable()
+export class AuthGuard implements CanActivate {
+  constructor(private router: Router, private authService: AppAuthService) {}
+
+  canActivate(_next: ActivatedRouteSnapshot, _state: RouterStateSnapshot): Observable<boolean> {
```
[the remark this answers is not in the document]
yenobi: fixed

src/app/login/login-page/login-page.component.html
yenobi: fixed

src/app/login/login-page/login-page.component.html
yenobi: fixed

src/app/login/login-page/login-page.component.ts
yenobi: updated this

src/app/login/login-page/login-page.component.html
yenobi: i have checked the examples in ui-ang and you are right, it is just my inner feel of how it is better.
fixed

santam85:
Why are we replacing the other auth library? The ultimate goal is NOT to maintain an auth library ourselves and to ONLY support OIDC auth (and not username/password fields within the app).
Please make it work with the 3rd party library against a running identity server (which it was last time I checked).

georgejolema (approved):
Very good job!

santam85 (approved):

src/app/services/app-auth.service.ts
```diff
@@ -1,71 +1,158 @@
...
-        map((user) => this.currentUser$$.next(user as User)),
-        map(() => this.isAuthenticated$$.next(true)),
-      )
-      .subscribe(() => this.redirectTo(redirectUri));
+  public runInitialLoginSequence(): Promise<void> {
+    // 0. LOAD CONFIG:
+    // First we have to check to see how the IdServer is
+    // currently configured:
+    return (
+      this.oauthService
```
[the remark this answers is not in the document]
yenobi: I have tried, but for some reason with calling only `loadDiscoveryDocumentAndTryLogin` the requests for auth update will go infinitely. I have tried to find the reason for such behavior, but failed and decided to go this way.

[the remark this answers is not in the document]
yash-backbase: Wouldn't something like this have worked?

```
    // configure auth code flow
    oAuthService.configure(authCodeFlowConfig);
    // load discovery document and try login attempt
    oAuthService.loadDiscoveryDocumentAndTryLogin()
      .then(() => {
        // if user already not logged in i.e has access token, initiate auth code flow
        if (oAuthService.hasValidAccessToken()) {
          // do nothing
          // ....
        } else {
          oAuthService.initCodeFlow();
        }
      });
```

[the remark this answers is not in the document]
santam85: Yeah that's what I meant! I'll try playing with it on master...

PR #15, opened 2021-09-26
First implementation

tcorral:
On this first implementation I have implemented the journey content component and the system to store the templates so they can be used anywhere on the application.

It's still pending the connector to a real BE, now it's just making use of a JSON BIN account.

It's necessary to implement the key per each journey content and this key must be the one used to set and provide the content id to be rendered.


santam85:
Is there a reason why the built version of the API client are committed to git? Would be easier to review the code without all of those

PR #17, opened 2021-10-01
Simpler authentication

santam85:
Remove some of the complexity to make auth flow more understandable.
Removed silent refresh as we have refresh tokens.


kirillsud (changes requested):
Well done. I like it!

One problem is why I didn't approve: tests are failing because of missing `app-auth.service`

src/app/guards/auth.guard.ts
```diff
@@ -1,18 +1,17 @@
...
-  canActivate(): Observable<boolean> {
-    return this.authService.isDoneLoading$.pipe(
-      filter((isDone) => isDone),
-      switchMap((_) => this.authService.isAuthenticated$),
-      tap((isAuthenticated) => isAuthenticated || this.authService.login()),
-    );
+  constructor(private oAuthService: OAuthService) {
   }
+
+  canActivate = () => {
```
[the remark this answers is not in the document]
santam85: not really, restored previous one

src/app/guards/auth.guard.ts
```diff
@@ -7,7 +7,7 @@ export class AuthGuard implements CanActivate {
   constructor(private oAuthService: OAuthService) {
   }
 
-  canActivate = () => {
+  canActivate(): boolean {
```
yenobi, line 10: ❤️

src/app/app.module.ts
```diff
@@ -8,7 +8,7 @@ import { AvatarModule, DropdownMenuModule, IconModule, LayoutModule, LogoModule
 import { EntitlementsModule } from '@backbase/foundation-ang/entitlements';
 import { HttpClientModule } from '@angular/common/http';
 import { AuthGuard } from './guards/auth.guard';
-import { WebSdkConfig, WebSdkModule } from '@backbase/foundation-ang/web-sdk';
+import { WebSdkModule } from '@backbase/foundation-ang/web-sdk';
```
igord, line 11: Why do we even need `WebSdkModule`?

PR #19, opened 2021-11-11
wf-411: Add feature to switch locale and add docker environment variables

georgejolema:


src/environments/environment.ts
```diff
@@ -45,21 +21,37 @@ const mockProviders: Provider[] = [
   },
 ];
 
+const pageConfig = {
```
santam85, line 24: I find the pageConfig object could be confusing things, and make people think we want to keep the web-sdk module and the lean page around

docker-compose.yml
```diff
@@ -0,0 +1,22 @@
...
+      - target: 4200
+        published: 4200
+    environment:
+      PROTOCOL: 'http:'
+      HOSTNAME: localhost
+      PORT: 4200
+      PATHNAME: /
+      API_ROOT: https://app.stable.retail.backbasecloud.com/api
+      AUTH_URL: https://identity-latest-universal.retail.backbase.eu/auth/realms/backbase
+      AUTH_REALM: backbase
```
santam85, line 17: Realms are a Keycloak specific concept, we don't need to make it explicit when using the full OIDC compliant auth method, this information is embedded in the AUTH_URL

src/environments/environment.prod.ts
```diff
@@ -1,20 +1,49 @@
+import { HTTP_INTERCEPTORS } from '@angular/common/http';
 import { AuthConfig } from 'angular-oauth2-oidc';
+import { TransactionsInterceptor } from '../app/interceptors/transactions.interceptor';
+import { AccountsInterceptor } from '../app/interceptors/accounts-interceptor';
+
+const pageConfig = {
+  locales: '${LOCALES}'.split(','),
```
santam85, line 7: Isn't there a way to retrieve the list of locales from Angular injection tokens? Did we look at Angular CLI docs/issues? Would be a nice candidate for a PR to the Angular project!

georgejolema: I will be checking if I can contribute. In the meantime I will leave this like that for the sake of stability and demo

docker-compose.yml
georgejolema: removed, thanks

src/environments/environment.ts
georgejolema: I removed the pageConfig completely

santam85 (approved):

yenobi:
What do you think about uniting the locale selector and localesCatalog to the module to have better separation of concepts?

default.conf.template
```diff
@@ -0,0 +1,97 @@
+# http://nginx.org/en/docs/http/ngx_http_headers_module.html
```
yenobi, line 1: I do not understand the purpose of the custom config, which is the same as we have in web-base.
Can you explain, what I am missing, please?

georgejolema: I left this for now here until next version of web-base gets released with the fixes so that we can demo or have others test. If web-base gets released before this I will remove it

georgejolema: now 0.0.12 is released so I removed it from here

README.md
```diff
@@ -38,6 +38,10 @@ Run `ng test` to execute the unit tests via [Karma](https://karma-runner.github.
 
 Run `ng e2e` to execute the end-to-end tests via [Protractor](http://www.protractortest.org/).
 
+## Running with docker
+
+Run `ng build --configuration production` and then `docker-compose up` to startup the docker container with the application
```
yenobi, line 43: Not crutial, but you can add additional npm script `"start:docker:compose": "ng build --configuration production && docker-compose up"`.

src/app/app.component.ts
```diff
@@ -1,7 +1,9 @@
-import { Component } from '@angular/core';
+import { Component, Inject, OnInit } from '@angular/core';
 import { LayoutService } from '@backbase/ui-ang/layout';
 import { triplets } from './services/entitlementsTriplets';
 import { OAuthService } from 'angular-oauth2-oidc';
+import { PageConfig, PAGE_CONFIG, SetLocale, SET_LOCALE } from '@backbase/foundation-ang/web-sdk';
```
yenobi, line 5: Unused imports, can be removed.

src/app/locale-selector/locales.service.ts
```diff
@@ -0,0 +1,18 @@
+import { Inject, Injectable, InjectionToken, LOCALE_ID } from "@angular/core";
+import { CookieService } from "ngx-cookie-service";
+export const LOCALES_LIST = new InjectionToken<Array<string>>('List of locales available');
+
+@Injectable()
+export class LocalesService {
+  setLocaleCookie(value: string) {
+    this.cookie.set('bb-locale', value);
```
yenobi, line 8: There is the only place, where we use the method from external library and according to the source code it can be changed to `this.document.cookie =  encodeURIComponent(name) + '=' + encodeURIComponent(value) + ';';`.
Do you think it is worth to have it?

README.md
georgejolema: sure I plan to make this nice in this ticket https://backbase.atlassian.net/browse/WF-408 I will make a pr today or tomorrow to do this and other changes

src/app/locale-selector/locale-selector.module.ts
```diff
@@ -0,0 +1,24 @@
...
+import { CommonModule } from "@angular/common";
+
+@NgModule({
+  declarations: [LocaleSelectorComponent],
+  imports: [DropdownSingleSelectModule, FormsModule, CommonModule],
+  exports: [LocaleSelectorComponent],
+  providers: [ LocalesService ],
+})
+export class LocaleSelectorModule {
+  static forRoot(config: { locales: Array<string> } ): ModuleWithProviders<LocaleSelectorModule> {
```
kirillsud, line 15: Do we really need that config for a demonstration locale swither?

src/app/locale-selector/locales.service.ts
kirillsud: Maybe we can use local storage?

src/environments/environment.prod.ts
```diff
@@ -1,20 +1,34 @@
...
 };
 
 export const authConfig: AuthConfig = {
   // Url of the Identity Provider
-  issuer: '${ISSUER}',
+  issuer: '${AUTH_URL}',
 
   // URL of the SPA to redirect the user to after login
-  redirectUri: window.location.origin + '/transactions',
+  redirectUri: '${PROTOCOL}//${HOSTNAME}:${PORT}${PATHNAME}${AUTH_LANDING_PAGE}',
```
kirillsud, line 28: Do you think that this environment will be applied only for docker execution? Not for a local run?

src/app/locale-selector/locales.service.ts
georgejolema: it can be worthy but maybe in the future if we see there are more cookies management. But for now I followed your suggestion and removed the cookie service and made it as you said

src/app/app.component.ts
georgejolema: fixed

src/app/locale-selector/locales.service.ts
georgejolema: @kirillsud this should use the cookies because nginx expects that `bb-locale` is present to route to the right locale

src/environments/environment.prod.ts
georgejolema: when we do `ng serve` it uses `environment.ts` but in the case of docker it will use production mode and take `environment.prod.ts` do you think it would be worth to have something like `environment.docker.ts`?

src/app/locale-selector/locale-selector.module.ts
georgejolema: I just wanted to avoid using variables from environment.ts file directly and also wanted to keep that provider inside the locale selector module.

src/environments/environment.prod.ts
kirillsud: I thought about that, but I think it's okay to leave it as is

yenobi (approved):

kirillsud (approved):

src/environments/environment.prod.ts
santam85: I'd revert this one ideed, as using window.location.origin is robust to using reverse proxies in between too...

PR #21, opened 2021-11-30
fix realms; update docker file to use latest web-base verison

cheshirsky:


cheshirsky:
Not sure if you guys want to capture these. If not- feel free to decline :-) These were issues I have to deal with while trying to deploy on BaaS, might be irrelevant for the purposes of the golden-sample itself. It's more to inform you about these.

santam85 (approved):

PR #22, opened 2021-12-01
CXCLOUD-1512: BASE_HREF implementation

echonax:
updated the environment.prod and index.html files to reflect BASE_HREF changes


echonax:
This is also related to: https://backbase.atlassian.net/browse/WF-540

src/index.prod.html
```diff
@@ -0,0 +1,13 @@
+<!doctype html>
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <title>GoldenSampleApp</title>
+  <base href="${BASE_HREF}">
```
santam85, line 6: there's no need to do it this way, you can simply pass a base-href in the build options:

```
 "production": {
              "baseHref": "whatever",
```

santam85 (changes requested):

src/environments/environment.prod.ts
```diff
@@ -25,7 +25,7 @@ export const authConfig: AuthConfig = {
   issuer: '${AUTH_URL}/realms/backbase',
 
   // URL of the SPA to redirect the user to after login
-  redirectUri: '${PROTOCOL}//${HOSTNAME}:${PORT}${PATHNAME}${AUTH_LANDING_PAGE}',
+  redirectUri: '${PROTOCOL}//${HOSTNAME}:${PORT}${BASE_HREF}${PATHNAME}${AUTH_LANDING_PAGE}',
```
santam85, line 28: can you change this to use `window.location` while you are at it? That way it does not need any interpolation

src/index.prod.html
echonax: but how will that be reflected to the html when the user changes the env variable?

santam85: If you put `"baseHref": "${BASE_HREF}",`, the resulting index.html would look exactly like the one you have now! The value would be replaced when the container starts.

echonax: I'm probably misunderstanding something so just to clarify: where should I be putting this 
```
"production": { 
      "baseHref": "whatever",
```
?
I presumed angular.json; but that won't work I believe since it's not part of the statics that we're putting into the docker image.

santam85: try putting that in the angular.json and running a build, then check the index.html in the dist folder

src/environments/environment.prod.ts
```diff
@@ -25,7 +25,7 @@ export const authConfig: AuthConfig = {
   issuer: '${AUTH_URL}/realms/backbase',
 
   // URL of the SPA to redirect the user to after login
-  redirectUri: '${PROTOCOL}//${HOSTNAME}:${PORT}${PATHNAME}${AUTH_LANDING_PAGE}',
+  redirectUri: window.location.href + 'transactions',
```
georgejolema, line 28: this change will break docker implementation. What is the reason for that?

src/index.prod.html
echonax: Oh I see.. I thought you thought it was going to be changed in angular.json. Resolving now.

src/environments/environment.prod.ts
echonax: Umm check the previous comment for the background. Lemme actually double-test this; you might be right

georgejolema: this `${PROTOCOL}//${HOSTNAME}:${PORT}${PATHNAME}${AUTH_LANDING_PAGE}` are env variables coming from web-base:  https://github.com/Backbase/golden-sample-app/blob/master/docker-compose.yml#L13

santam85: Also, the variable names we added to the golden sample don't seem to match with our original guideline:
https://backbase.atlassian.net/wiki/spaces/DE/pages/3020391409/How+to+dockerize+an+Angular+app

AUTH_LANDING_PAGE and AUTH_REDIRECT_PAGE don't need to be configurable per environment, as they are relative to the app, so we can remove those variables too.

echonax: tested btw; it works

santam85 (approved):
Perfect

PR #23, opened 2021-12-02
Bugfix/token refresh

jordydejong:
The repo in its current state was not refreshing the token. So after x minutes, new API calls would be failing.

When adding `oAuthService.setupAutomaticSilentRefresh();` a refresh was triggered, however a complete refresh of the page was done

Only after removing `useSilentRefresh: true` all worked as expected. But, to be honest, not 100% sure why, could not find documentation for this behaviour.

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -67,7 +67,11 @@ const buildEntitlementsByUser =
       provide: APP_INITIALIZER,
       multi: true,
       deps: [ OAuthService ],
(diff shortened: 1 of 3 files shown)


santam85 (approved):
It was just a relic of the implicit flow based example. Your fix is the correct implementation, also according to the docs (https://manfredsteyer.github.io/angular-oauth2-oidc/docs/additional-documentation/token-refresh.html)

PR #25, opened 2021-12-08
Example of routing configuration for a view with modal component

jesus-crysist:
In some journeys, it will be required to have a modal dialog displayed on top of the main view. That modal dialog, unless it is a prompt dialog, will probably be new view of the journey and thus should contain its own route configuration.

This PR is providing an example of how we can use named router outlet to add component with modal as a view.

Main view of the journey does not contain any data and is there just to provide preview of how the dialog will be displayed on top of it. The form in modal dialog does not contain any validations on purpose, which is not the topic of this example.

_Note:_ The PR also contains a fix for issue of Positive Pay journey with Entitlements.


src/app/ach-positive-pay/ach-positive-pay-journey-bundle.module.ts
```diff
@@ -0,0 +1,21 @@
+import { NgModule } from '@angular/core';
+import { AchPositivePayJourneyModule } from 'ach-positive-pay-journey';
+import { CONDITIONS } from '@backbase/foundation-ang/web-sdk';
+import { buildEntitlementsByUser, triplets } from '../services/entitlementsTriplets';
+
+@NgModule({
+  imports: [AchPositivePayJourneyModule.forRoot()],
+  providers: [
+    {
+      provide: CONDITIONS,
```
georgejolema, line 10: there is no more need to provide this anymore. Now you can add your entitlements here https://github.com/Backbase/golden-sample-app/blob/master/src/app/interceptors/entitlements-interceptor.ts

src/app/app.component.ts
```diff
@@ -25,7 +21,7 @@ export class AppComponent {
   }
 
   focusMainContainer(event: MouseEvent) {
-    const element = event.view?.window?.document?.querySelector('[role=\'main\']') as HTMLElement | undefined;
+    const element = event.view?.window?.document?.querySelector("[role='main']") as HTMLElement | undefined;
```
georgejolema, line 24: lint complains with the double quotes

jesus-crysist: Fixed.

src/app/ach-positive-pay/ach-positive-pay-journey-bundle.module.ts
jesus-crysist: Updated.

santam85 (approved):
LGTM

georgejolema (approved):

PR #26, opened 2021-12-15
Nojira/user identity sandbox env

yenobi:


Farzannajipour (approved):

src/environments/environment.ts
```diff
@@ -30,7 +30,7 @@ export const environment = {
 
 export const authConfig: AuthConfig = {
   // Url of the Identity Provider
-  issuer: 'https://identity.ret-us-l.rnd.live.backbaseservices.com/auth/realms/backbase',
+  issuer: 'https://identity.dev.sdbx.live.backbaseservices.com/auth/realms/retail',
```
georgejolema, line 33: can you also check this https://github.com/Backbase/golden-sample-app/blob/master/docker-compose.yml#L16 and update if possible

src/environments/type.ts
```diff
@@ -0,0 +1,9 @@
+import { Provider } from "@angular/core";
+
+export interface Environment {
+	production: boolean,
+	apiRoot: string,
+	mockProviders: Provider[],
+	locales: string[],
+	apiURL?: string,
```
santam85, line 8: what's the difference between apiRoot and apiURL?

santam85 (changes requested):

georgejolema (approved):

src/environments/type.ts
georgejolema: it should only be apiRoot now

yenobi: removed `apiURL` from type and the actual value to avoid confusion

santam85 (approved):

PR #38, opened 2022-04-13
Add example for journey extension

jordydejong:
> We want to provide customers the option to extend journey views, without the need to replace the entire view. For instance, when adding a small piece of html somewhere in the journey.

This is an example for the above, an extension slot replacement. 

The solution is based on an earlier POC by Armin

### DIFF
--- a/apps/golden-sample-app/src/environments/environment.ts
+++ b/apps/golden-sample-app/src/environments/environment.ts
@@ -55,8 +55,6 @@ export const authConfig: AuthConfig = {
   showDebugInformation: true,
 
   logoutUrl: document.location.origin + '/logout',
-  useSilentRefresh: true,
-  silentRefreshTimeout: 5000,
 };
 
 /*
--- a/libs/transactions/src/lib/components/transaction-item/transaction-item.component.html
+++ b/libs/transactions/src/lib/components/transaction-item/transaction-item.component.html
@@ -18,11 +18,7 @@
       *ngIf="additionsDetailsTpl"
       data-slot="additionsDetails"
       [ngTemplateOutlet]="additionsDetailsTpl"
-      [ngTemplateOutletContext]="{
-        additions: transaction.additions,
-        merchant: transaction.merchant,
-        counterPartyAccountNumber: transaction.counterPartyAccountNumber
-      }"
+      [ngTemplateOutletContext]="additionsDetailsContext"
     ></ng-container>
   </span>
 </span>
--- a/libs/transactions/src/lib/components/transaction-item/transaction-item.component.html
+++ b/libs/transactions/src/lib/components/transaction-item/transaction-item.component.html
@@ -18,7 +18,11 @@
       *ngIf="additionsDetailsTpl"
       data-slot="additionsDetails"
       [ngTemplateOutlet]="additionsDetailsTpl"
-      [ngTemplateOutletContext]="{ $implicit: transaction }"
+      [ngTemplateOutletContext]="{
+        additions: transaction.additions,
+        merchant: transaction.merchant,
+        counterPartyAccountNumber: transaction.counterPartyAccountNumber
+      }"
     ></ng-container>
   </span>
 </span>
--- a/libs/transactions/src/lib/components/transaction-item/transaction-item.component.html
+++ b/libs/transactions/src/lib/components/transaction-item/transaction-item.component.html
@@ -12,9 +12,8 @@
   <span class="transactions-item__counterParty">{{
     transaction.counterPartyName
   }}</span>
-  <span class="transactions-item__transactionType">{{ transaction.type }}</span>
-
-  <span class="transactions-item__additionalDetails">
+  <span class="transactions-item__transactionType">
+    {{ transaction.type }}
     <ng-container
       *ngIf="additionsDetailsTpl"
       data-slot="additionsDetails"
@@ -22,7 +21,6 @@
       [ngTemplateOutletContext]="{ $implicit: transaction }"
     ></ng-container>
   </span>
-
 </span>
 <span
   class="transactions-item__amount"
--- a/libs/transactions/src/lib/components/transaction-item/transaction-item.component.html
+++ b/libs/transactions/src/lib/components/transaction-item/transaction-item.component.html
@@ -9,10 +9,20 @@
   transaction.valueDate | date: 'MMM. d'
 }}</span>
 <span class="transactions-item__description">
-  <span class="transactions-item__merchant">{{
-    transaction.merchant?.name
+  <span class="transactions-item__counterParty">{{
+    transaction.counterPartyName
   }}</span>
   <span class="transactions-item__transactionType">{{ transaction.type }}</span>
+
+  <span class="transactions-item__additionalDetails">
+    <ng-container
+      *ngIf="additionsDetailsTpl"
+      data-slot="additionsDetails"
+      [ngTemplateOutlet]="additionsDetailsTpl"
+      [ngTemplateOutletContext]="{ $implicit: transaction }"
+    ></ng-container>
+  </span>
+
 </span>
 <span
   class="transactions-item__amount"
--- a/libs/transactions/src/lib/components/transaction-item/transaction-item.component.ts
+++ b/libs/transactions/src/lib/components/transaction-item/transaction-item.component.ts
@@ -4,8 +4,10 @@ import {
   Input,
   OnChanges,
   SimpleChanges,
+  TemplateRef,
 } from '@angular/core';
 import { TransactionItem } from '@backbase/data-ang/transactions';
+import { TransactionsJourneyConfiguration } from '../../services/transactions-journey-config.service';
 
 @Component({
   selector: 'bb-transaction-item',
@@ -17,9 +19,17 @@ export class TransactionItemComponent implements OnChanges {
   @Input() transaction!: TransactionItem;
   public amount = 0;
   public isAmountPositive = true;
+  public additionsDetailsTpl: TemplateRef<any> | undefined;
+  
+  constructor(private readonly configService: TransactionsJourneyConfiguration) {
+    this.additionsDetailsTpl = this.configService.additionalDetailsTpl;
 
+  }
+  
   ngOnChanges(changes: SimpleChanges): void {
     if (changes['transaction']) {
+    console.warn('transaction', this.transaction);
+
       this.amount = Number(
         this.transaction.transactionAmountCurrency.amount ?? 0
       );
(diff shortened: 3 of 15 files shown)


libs/transactions/src/lib/components/transaction-item/transaction-item.component.html
```diff
@@ -9,10 +9,20 @@
   transaction.valueDate | date: 'MMM. d'
 }}</span>
 <span class="transactions-item__description">
-  <span class="transactions-item__merchant">{{
-    transaction.merchant?.name
```
jordydejong, line 13: Currently there is no merchant visible, since it is not available in the retrieved data.

libs/transactions/src/lib/services/transactions-journey-config.service.ts
```diff
@@ -1,7 +1,8 @@
-import { Injectable } from '@angular/core';
+import { Injectable, TemplateRef } from '@angular/core';
 
 @Injectable()
 export class TransactionsJourneyConfiguration {
   pageSize = 20;
   slimMode = true;
-}
+  additionalDetailsTpl: TemplateRef<any> | undefined = undefined;
```
jordydejong, line 7: Should we add a type for the context the template uses for rendering?
Or is this overkill, and will `any` suffice?

libs/transactions/src/lib/transactions-journey.component.ts
```diff
@@ -0,0 +1,17 @@
+import { AfterContentChecked, Component, ContentChild } from '@angular/core';
+import { TransactionAdditionalDetailsTemplateDirective } from './directives/transaction-additional-details.directive';
+import { TransactionsJourneyConfiguration } from './services/transactions-journey-config.service';
+
+@Component({
+  selector: 'bb-transactions-journey',
+  templateUrl: 'transactions-journey.component.html',
+})
+export class TransactionsJourneyComponent implements AfterContentChecked {
```
jordydejong, line 9: It seems currently it is not always common that a journey exports its main journey component, or even that it has one. In this proposal it would be a pre-requisite in order to enable providing templates from the outside

apps/golden-sample-app/src/environments/environment.ts
```diff
@@ -55,8 +55,6 @@ export const authConfig: AuthConfig = {
   showDebugInformation: true,
 
   logoutUrl: document.location.origin + '/logout',
-  useSilentRefresh: true,
-  silentRefreshTimeout: 5000,
```
jordydejong, line 59: This configuration was re-added, but (at least for me) causes automatic refresh to not work anymore. So that means API calls start failing after a couple of minutes.

libs/transactions/src/lib/services/transactions-journey-config.service.ts
jordydejong: added typing in new commit

libs/transactions/src/lib/directives/transaction-additional-details.directive.ts
```diff
@@ -1,7 +1,15 @@
 import { Directive, TemplateRef } from "@angular/core";
+import { TransactionItem } from "@backbase/data-ang/transactions";
 
+export type AdditionalDetailsContext = {
+  $implicit: TransactionItem
```
jordydejong, line 5: we could either combine a key of `$implicit` with `let-transaction` in the template

or `transaction: TransactionItem` in combination with `let-transaction="transaction"`

libs/transactions/src/lib/directives/transaction-additional-details.directive.ts
```diff
@@ -1,7 +1,15 @@
...
+}
 
 @Directive({selector: 'ng-template[bbTransactionAdditionalDetails]'})
 export class TransactionAdditionalDetailsTemplateDirective {
-  constructor(public templateRef: TemplateRef<any>) {}
+  constructor(public templateRef: TemplateRef<AdditionalDetailsContext>) {}
+
+  static ngTemplateContextGuard(dir: TransactionAdditionalDetailsTemplateDirective, ctx: unknown): ctx is AdditionalDetailsContext {
```
jordydejong, line 12: this adds autocomplete / compile time errors for developers in the template when implementing this  extension

if we think this is overkill (I don't) we could just type the templateRef with `any` as the generic

santam85: I like this, it makes it easy to detect breaking changes to the public APIs

libs/transactions/src/lib/services/transactions-journey-config.service.ts
```diff
@@ -1,8 +1,9 @@
 import { Injectable, TemplateRef } from '@angular/core';
+import { AdditionalDetailsContext } from '../directives/transaction-additional-details.directive';
 
 @Injectable()
 export class TransactionsJourneyConfiguration {
   pageSize = 20;
   slimMode = true;
-  additionalDetailsTpl: TemplateRef<any> | undefined = undefined;
+  additionalDetailsTpl: TemplateRef<AdditionalDetailsContext> | undefined = undefined;
```
santam85, line 8: This is all an engineer would need to do to override the default null template, right?

jordydejong: The developer should define the interface for the context in the directive, and write a guard for it

And they should import this (like here) in the config service to correctly type the templates stored there

libs/transactions/src/lib/components/transaction-item/transaction-item.component.ts
```diff
@@ -20,7 +18,7 @@ export class TransactionItemComponent implements OnChanges {
   @Input() transaction!: TransactionItem;
   public amount = 0;
   public isAmountPositive = true;
-  public additionsDetailsTpl: TemplateRef<AdditionalDetailsContext> | undefined;
```
jordydejong: I suggest not to use `public additionsDetailsTpl: TemplateRef<AdditionalDetailsContext> | undefined;` here, since

1. it would create yet another place where we have to import this context's interface
2.this type is correctly inferred from the (typed) config service

libs/transactions/src/lib/components/transaction-item/transaction-item.component.html
```diff
@@ -9,10 +9,20 @@
...
+  <span class="transactions-item__counterParty">{{
+    transaction.counterPartyName
   }}</span>
   <span class="transactions-item__transactionType">{{ transaction.type }}</span>
+
+  <span class="transactions-item__additionalDetails">
```
yenobi, line 17: I want to challenge this in the example.
First of all because of the span compared to the div.
Secondly, it will be needed to override styles from the inside of the slot if they won't match the customers' theme.
Thirdly, if the styles will be provided on the level where we define the content of the addition slot, then it can serve as an example of how to do this.

yenobi (commented):
Also, looks like the directive and the type are missing in the public API (aka index.ts right now).

Overall, i like the proposal and working example with all of the comments. Good work @jordydejong

libs/transactions/src/lib/components/transaction-item/transaction-item.component.html
```diff
@@ -9,10 +9,20 @@
...
+    <ng-container
+      *ngIf="additionsDetailsTpl"
+      data-slot="additionsDetails"
```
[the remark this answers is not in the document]
jordydejong: Agree, but since this repo doesn't have any e2e tests yet, I propose to pick this up separately.

jordydejong:
> Also, looks like the directive and the type are missing in the public API (aka index.ts right now).
> 
> Overall, i like the proposal and working example with all of the comments. Good work @jordydejong

Thanks for the feedback.
I am not yet very familiar on what a journey is supposed to export in its public API. No problem adding that of course, but for what scenario would that be necessary? Will people use this interface or directive directly, without importing the entire module?

libs/transactions/src/lib/components/transaction-item/transaction-item.component.html
jordydejong: > First of all because of the span compared to the div.
What exactly do you mean with this?

I will investigate / add an example for overriding styles

jordydejong: Journey no longer adds wrapper + styling to the extension anymore, expects app/implementation to style the extension

Moved the bundle component to separate ts/html/scss files as part of this.

santam85 (approved):

libs/transactions/src/lib/directives/transaction-additional-details.directive.ts
cheshirsky: I would suggest to narrow down the example to the additions field in a transaction item (if exists) and avoid passing entire object in

libs/transactions/src/lib/directives/transaction-additional-details.directive.ts
cheshirsky: I would suggest to add a simple test to the example

libs/transactions/src/lib/components/transaction-item/transaction-item.component.html
yenobi: Regarding the `span` and `div` difference, I mean two things:
- that the first one is an inline-block and the second is a block
- the content model difference between the tags. `span` belongs to the `phrasing content` and for this example, it works perfectly, but as a generic example the `flow content` of the div would suit more. And because it might be missed from the reader's perspective I would vote to have the `div` if we want to have the wrapper for the additional slot.

yenobi (approved):

libs/transactions/src/lib/directives/transaction-additional-details.directive.ts
jordydejong: @cheshirsky  Sure I can do that. So it's not a common scenario that, when implementing a journey, we want to add one of the existing properties to the view?

jordydejong: Not sure if it makes much sense to type the context in that case, since the additions are very loosely typed as `[key: string]: string;`

jordydejong:
I have made the following changes

- Added tests for 'capturing' the template in the journey, and rendering it in the transaction item component

- Instead of providing the full transaction item object to the extension template, the journey now specifically decides what specific subset of data is available for the template to use. This includes the additional data. Hopefully this takes away some of the concern regarding potential API changes impacting the extensions.

- As well as one of these transaction properties, the app now also tries to render a custom addition. This specific addition does not exist in our transaction item (in fact, the sandbox API doesn't return any additiosn at all) but should demo how to render a custom addition field.

If approved I will update the confluence page accordingly. Future improvements to this example could be to actually render "real" addition data coming from the API, adding fields to forms, or displaying custom functionality like a print icon below the transaction list.

PR #40, opened 2022-04-22
test implementation ng-dynamic-component

jordydejong:
alternative to https://github.com/Backbase/golden-sample-app/pull/38/files that is currently in main

pros
- no directive
- no need for wrapper component around journey
- no need to add template to config service

cons:
- less type safe (?) at least in this implementation
- need to provide entire component for every addition, instead of only a template
- third party library


libs/transactions/src/lib/tokens/index.ts
```diff
@@ -0,0 +1 @@
+export * from './transaction-addition-details.token';
```
yenobi, line 1: Do not see any purpose for this barrel

yenobi (commented):
I am not a fan of using external libraries for the use cases when there is an OOTB possibility tbh. Looks like the support of the library is not very good from a quick glance (seasonal merges with a gap of half a year, long-living issues, and so on).

But I like the idea of injection token implementation.

UPD: I did not check the description before the code itself. Agree that the advantages look good, it might be a good idea to think about the same way of implementation but limit the third party effect with own OSS implementation.

PR #41, opened 2022-04-25
 test: ngComponentOutlet with custom injector

jordydejong:
same as https://github.com/Backbase/golden-sample-app/pull/40, but using a custom injector instead of ng-dynamic-component

open issues
- currently things like *ngIf don't seem to be working in the provided template
- data  provided to extension component is not "dynamic"

pros (compared to other draft PR):
- no third party dependency
- data going from journey to app is properly typed

cons  (compared to other draft PR):
- more complicated to create an extension by journey developer
- the component supplied to the journey gets data by injection token instead of inputs, which makes it less of a "standard" component
- the solution in this PR currently only supplies the initial data/values, at the moment it is injected somewhere. So no updated values like an @input would do.  We could probably work around this by providing the extension component with an observable around the data.

the ng-dynamic-component works around these cons, it seems, by adding a lot of custom code and hacking into low level Angular properties and methods.


jordydejong:
closing this POC pull request, we can have potential further discussions in slack/confluence

PR #44, opened 2022-04-28
POC single route replacement

jordydejong:
POC for overriding one or more specific routes, instead of providing an entire new set of routing to the journey

This prevents copy/pasting the entire route config, and importing all routable components

 I assume this would make it less difficult to upgrade


santam85:
I like the idea, perhaps we can do it in a simpler way? Maybe just by leveraging the priority the router checks routes it is enough to just prepend the new ones?

jordydejong:
> I like the idea, perhaps we can do it in a simpler way? Maybe just by leveraging the priority the router checks routes it is enough to just prepend the new ones?

I think the advantage of this approach, is that any other config for that route is still applied and does not have to be exposed publicly. Entitlements, resolvers and so on.

PR #46, opened 2022-05-17
additions components via config

daiscog:
Alternative to having to override the routing to render dynamic view extensions.

Most of the legwork is done in an abstract Directive class which would be provided by foundation-ang:

```ts
export interface ExtensionComponent<ContextType> extends Partial<OnChanges & OnDestroy & OnInit> {
  set context(context: ContextType | undefined);
}

@Directive()
export abstract class ExtensionSlotDirective<ContextType> implements OnInit, OnDestroy, OnChanges {
  @Input() componentType: Type<ExtensionComponent<ContextType>> | undefined;
  @Input() context: ContextType | undefined;

  private componentRef: (ComponentRef<ExtensionComponent<ContextType>>) | undefined;

  constructor(
    private readonly vcRef: ViewContainerRef,
    private readonly cfr: ComponentFactoryResolver
  ) {}

  ngOnChanges(changes: SimpleChanges): void {
    const componentInstance = this.componentRef?.instance;
    if (changes.context && componentInstance?.ngOnChanges) {
      componentInstance.context = this.context;
      componentInstance.ngOnChanges({
        context: changes.context
      });
    }
  }

  ngOnDestroy(): void {
    this.vcRef.clear();
    this.componentRef?.destroy();
  }

  ngOnInit(): void {
    if (!this.componentType) {
      throw new Error('componentType not defined');
    }
    // In Angular 13 we don't need the ComponentFactoryResolver and can just
    // pass this.componentType directly to this.vcRef.createComponent instead
    const factory = this.cfr.resolveComponentFactory(this.componentType);
    this.componentRef = this.vcRef.createComponent(factory);
    this.componentRef.instance.context = this.context;
  }
}
```

## Journey team effort

Journey teams would expose a type-safe config object to allow passing in the extension component type via the journey module forRoot function:

```ts
// using the generic ExtensionComponent interface defined in foundation-ang
export type TransactionAdditionalDetailsComponent = ExtensionComponent<TransactionItem>; 

export interface TransactionsJourneyExtensionsConfig {
  transactionItem?: Type<TransactionAdditionalDetailsComponent>;
  // other extensions slots here
}

export interface TransactionsJourneyModuleConfig {
  route?: Route;
  extensionSlots?: TransactionsJourneyExtensionsConfig
}

export class TransactionsJourneyModule {
  static forRoot(
    {route, extensionSlots}: TransactionsJourneyModuleConfig = {}
  ): ModuleWithProviders<TransactionsJourneyModule> {
    return {
      ngModule: TransactionsJourneyModule,
      providers: [
        provideRoutes([route || defaultRoute]),
        {
          provide: ΘTRANSACTION_EXTENSIONS_CONFIG,
          useValue: extensionSlots || {}
        }
      ],
    };
  }
}
```

Extension slot component config is then available in the native journey component like so:

```ts
export class TransactionItemComponent {
  //...
  public additionsDetails?: Type<ExtensionComponent<TransactionItem>>;

  constructor(
    @Inject(ΘTRANSACTION_EXTENSIONS_CONFIG) extensionsConfig: TransactionsJourneyExtensionsConfig,
  ) {
    this.additionsDetails = extensionsConfig.transactionItem;
  }

  //...
}
```

Journey would define a concrete extensions of the foundation-ang ExtensionSlotDirective, again using generics to state the type of data that will be passed to the component:

```ts
@Directive({
  selector: '[bbTransactionsItemAdditions]'
})
export class TransactionItemAdditonalDetailsDirective extends ExtensionSlotDirective<TransactionItem> {}
```

This would be used in a template like so:

```html
    <ng-container
      *ngIf="additionsDetails"
      bbTransactionsItemAdditions
      [componentType]="additionsDetails"
      [context]="transaction"
    >
    </ng-container>
```

## CS/Customer team effort

All the customer would need to do is make a component that implements the `TransactionAdditionalDetailsComponent` type:

```ts
@Component(...)
export class TransactionItemAdditionalDetailsComponent implements TransactionAdditionalDetailsComponent {
  @Input()
  context: TransactionItem | undefined;
}
```

And configure it in the Journey module:

```ts
@NgModule({
  imports: [
    TransactionsJourneyModule.forRoot({
      extensionSlots: {
        transactionItem: TransactionItemAdditionalDetailsComponent
      }
    })
  ],
  //...
})
export class TransactionsJourneyBundleModule {}
```

## Pros

* Typesafe extension component type
* Handles changes in the data
* No need to override routing or write a custom top-level journey component
* Very low level of effort for customers
* Reasonably low level of effort for RnD
* No need for customers to manually `provide` DI values

## Cons

* Requires some initial work to provide common legwork via foundation-ang


santam85 (approved):
Some conflicts to solve, but looks good to me! @tcorral

libs/transactions/src/lib/extension-slot.directive.ts
```diff
@@ -0,0 +1,61 @@
+/*
```
georgejolema, line 1: As I understand this would go in foundation-ang. My only concern is that we are trying to reduce (or remove if possible) our core libraries so that we have less things to maintain and customers are more independent from us

libs/transactions/src/lib/extensions/transaction-addition-details.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export type TransactionAdditionalDetailsComponent = ExtensionComponent<TransactionItem>;
+
+export interface TransactionsJourneyExtensionsConfig {
+  transactionItem?: Type<TransactionAdditionalDetailsComponent>;
+  // other extensions slots here
+}
+
+// Private injection token
+export const ΘTRANSACTION_EXTENSIONS_CONFIG = new InjectionToken<TransactionsJourneyExtensionsConfig>('TRANSACTION_EXTENSIONS_CONFIG');
```
jordydejong, line 13: is is common (for us) to mark variables private with `Θ` ?

libs/transactions/src/lib/extensions/transaction-addition-details.ts
```diff
@@ -0,0 +1,13 @@
+import {InjectionToken, Type} from "@angular/core";
+import { TransactionItem } from "@backbase/data-ang/transactions";
+import {ExtensionComponent} from "../extension-slot.directive";
+
+export type TransactionAdditionalDetailsComponent = ExtensionComponent<TransactionItem>;
```
jordydejong, line 5: As an example of a best practice, I think we shouldn't use the entire `TransactionItem` as the component's context, unnecessarily exposing the entire object as (basically) part of the public API

jordydejong:
First of all, great job. I tried to investigate something along these lines but became stuck with using deprecated Angular code.

I greatly prefer this approach from the applications' perspective, seems much easier (and more logical) to just provide the journey with a component to render, instead of a wrapper component.

I have some concerns about edge cases and upgradability of this solution, because it seems to be digging a bit deeper in Angular internals. I can't find any concrete issues though.

I do think we should either support this one or the other. So if we agree on this one, I would suggest updating community and confluence with this solution. And maybe mentioning in an R&D sync that we changed our minds ;)

libs/transactions/src/lib/extension-slot.directive.ts
daiscog: Yeah, the idea is the `ExtensionComponent` interface and the `ExtensionSlotDirective` abstract class would go somewhere central like in foundation-ang, so RnD teams don't have to duplicate the code in every journey.  This also reduces risk that they'll implement it incorrectly, and enables bugfixes, or Angular API changes like the removal of the use if `ComponentFactoryResolver` (which was deprecated in Ang 13) in one place.

libs/transactions/src/lib/extensions/transaction-addition-details.ts
daiscog: Some teams in BB do it, but not many. It's a convention I like.

libs/transactions/src/lib/extensions/transaction-addition-details.ts
daiscog: Yeah, that's easily dealt with.  Will refactor.

libs/transactions/src/lib/extensions/transaction-addition-details.ts
jordydejong: i don't have a (very) strong opinion on this, besides preferring consistency

daiscog: Tbh, I just did it here to emphasise this injection token does not need to be exported and customers do not need to manually create a provider for it as it's all handled internally to the journey in a type-safe way.

kirillsud (approved):
It's a great improvement for the previous solution!

santam85:
[] Some conflitcs pending

libs/transactions/src/lib/components/transaction-item/transaction-item.component.ts
```diff
@@ -46,3 +72,8 @@ export class TransactionItemComponent implements OnChanges {
     }
   }
 }
+
+@Directive({
+  selector: '[bbTransactionsItemAdditions]'
+})
+export class TransactionItemAdditonalDetailsDirective extends ExtensionSlotDirective<TransactionItem> {}
```
jordydejong, line 79: I assume  this type should be `TransactionAdditionalDetailsContext` now, instead of TransactionItem

jordydejong, line 79: I assume  this type should be `TransactionAdditionalDetailsContext` now, instead of TransactionItem

libs/transactions/src/lib/extensions/transaction-addition-details.ts
```diff
@@ -0,0 +1,14 @@
+import {InjectionToken, Type} from "@angular/core";
+import { TransactionItem } from "@backbase/data-ang/transactions";
+import {ExtensionComponent} from "../extension-slot.directive";
+
+export type TransactionAdditionalDetailsContext = Pick<TransactionItem, 'additions' | 'counterPartyAccountNumber' | 'merchant'>;
+export type TransactionAdditionalDetailsComponent = ExtensionComponent<TransactionAdditionalDetailsContext>;
+
+export interface TransactionsJourneyExtensionsConfig {
```
jordydejong, line 8: since this config is shared, containing all extension slots, I assume it doesn't belong in this (extension specific) file

jordydejong, line 8: since this config is shared, containing all extension slots, I assume it doesn't belong in this (extension specific) file

jordydejong, line 8: since this config is shared, containing all extension slots, I assume it doesn't belong in this (extension specific) file

jordydejong (changes requested):

libs/transactions/src/lib/components/transaction-item/transaction-item.component.ts
jordydejong, line 79: I assume this type should be TransactionAdditionalDetailsContext now, instead of TransactionItem, and exampleChanges$ should also not return the entire transaction item

jordydejong (changes requested):

libs/transactions/src/index.ts
```diff
@@ -1,8 +1,16 @@
 export * from './lib/services/transactions-journey-config.service';
 export * from './lib/transactions-journey.module';
+export {
+  TransactionsJourneyExtensionsConfig,
+  TransactionAdditionalDetailsComponent,
+  TransactionAdditionalDetailsContext,
+  // don't export the private config injection token
+} from './lib/extensions';
```
jordydejong, line 8: is it not enough to only export TransactionAdditionalDetailsComponent?

apps/golden-sample-app/src/app/transactions/transaction-additional-details.component.html
```diff
@@ -0,0 +1,4 @@
+<span class="transaction-account-number">{{ context?.counterPartyAccountNumber }}</span>
+<span class="transaction-my-addition" *ngIf="context?.additions?.['myAddition']">
```
jordydejong, line 2: something like `context?.additions?.['myAddition']` looks a bit messy in my opinion, any suggestions for cleaning this up a bit?

daiscog (commented):
I've updated to ui-ang 7.5.0 and refactored accordingly

jordydejong (approved):

santam85:
LGTM! I'll let @cheshirsky have a last look and then merge!

package.json
```diff
@@ -39,16 +39,14 @@
     "@backbase/positive-pay-check-v1-client-ang": "^1.3.0",
     "@backbase/positive-pay-journey-ang": "^0.44.0",
     "@backbase/select-context-widget-ang": "^4.4.0",
-    "@backbase/ui-ang": "^6.51.0",
+    "@backbase/ui-ang": "^7.5.0",
     "@backbase/universal-ang": "^7.0.0",
-    "@ng-bootstrap/ng-bootstrap": "^9.1.3",
-    "@ng-select/ng-select": "^6.1.0",
```
cheshirsky, line 45: not sure why ng-select and bignumber are removed. Is it expected?

libs/transactions/src/lib/components/transaction-item/transaction-item.component.ts
```diff
@@ -16,22 +24,21 @@ import { ExtensionTemplatesService } from '../../services/extension-templates.se
...
-    return {
-      additions: this.transaction.additions,
-      merchant: this.transaction.merchant,
-      counterPartyAccountNumber: this.transaction.counterPartyAccountNumber
-    }
+  public additionsDetails?: Type<TransactionAdditionalDetailsComponent>;
+  public additionsDetailsContext?: TransactionAdditionalDetailsContext;
+
+  constructor(
+    @Inject(TRANSACTION_EXTENSIONS_CONFIG)
```
cheshirsky, line 36: Should it be `@Optional`?

libs/transactions/src/index.ts
```diff
@@ -1,11 +1,14 @@
-export * from './lib/transactions-journey.component';
-export * from './lib/directives/transaction-additional-details.directive';
-
 export * from './lib/services/transactions-journey-config.service';
```
cheshirsky, line 1: I do realise it's not a part of your change, but could you factor out exporting via `*`, it's dangerous from public API stability perspective and I don't think it promotes the best practices.

libs/transactions/src/lib/extensions/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './transaction-addition-details';
```
cheshirsky, line 1: I don't see the approach with an internal index files a lot, would it make sense to avoid using it, to make it consistent? Or opposite, use it then everywhere.

libs/transactions/src/lib/transactions-journey.module.ts
```diff
@@ -28,14 +33,18 @@ const defaultRoute: Route = {
   },
 };
 
+export interface TransactionsJourneyModuleConfig {
```
cheshirsky, line 36: I don't think I have seen it in the public API exports. Can you please check the PR via Microsoft API extractor, to make sure we have no forgotten exports?

apps/golden-sample-app/src/app/transactions/transaction-additional-details.component.html
```diff
@@ -0,0 +1,9 @@
+<span class="transaction-account-number">{{
+  context?.counterPartyAccountNumber
+}}</span>
+<span
+  class="transaction-my-addition"
+  *ngIf="context?.additions?.['myAddition']"
+>
+  My Addition: {{ context?.additions?.['myAddition'] }}
```
cheshirsky, line 8: Shall we advice this way of accessing data from inner levels? I would suggest handling it in the component code. This way it can look cleaner and we can demo setting defaults as well.

cheshirsky (approved):

package.json
santam85: Those became direct dependencies of ui-ang, rather than peer deps AFAIK

apps/golden-sample-app/src/app/transactions/transaction-additional-details.component.html
santam85: You mean the safe dereferencing op?

package.json
daiscog: Yes, ui-ang now brings these in as regular deps, so they no longer need to be explicitly installed.

That said, I'm not sure _why_ they are now regular dependencies instead of peerDependencies.

This can actually cause build issues if your app project declares incompatible versions in its root package.json, as ui-ang can end up getting its own copy of these deep inside node_modules that doesn't get processed by `ngcc`, then you end up with the horrible `importing a module that does not have a Θmod property` error.

If this happens, npm doesn't give you any warnings or errors.  But if they were peerDependencies, then npm would warn/error if incompatible versions are declared in the app package.json.

PR #48, opened 2022-06-22
Track clicks, navigation and custom events

tcorral:
**Do not merge this, this is just for demo purposes!!!!!**

On this example I have set two different listeners, one listening for clicks and another one listening for a custom event without adding anything custom and keeping it simple.

### DIFF
--- a/apps/golden-sample-app/src/app/app.component.ts
+++ b/apps/golden-sample-app/src/app/app.component.ts
@@ -22,7 +22,7 @@ export class AppComponent {
     // Listening for navigation events
     router.events.forEach((event) => {
       if(event instanceof NavigationEnd) {
-        (window as any).trackingEventsFake('change language:', event);
+        (window as any).trackingEventsFake('on navigate:', event);
       }
       // NavigationStart
       // NavigationEnd
--- a/apps/golden-sample-app/src/app/app.component.ts
+++ b/apps/golden-sample-app/src/app/app.component.ts
@@ -2,6 +2,7 @@ import { Component } from '@angular/core';
 import { LayoutService } from '@backbase/ui-ang/layout';
 import { triplets } from './services/entitlementsTriplets';
 import { OAuthService } from 'angular-oauth2-oidc';
+import { NavigationEnd, Router } from '@angular/router';
 
 @Component({
   selector: 'app-root',
@@ -14,9 +15,21 @@ export class AppComponent {
 
   constructor(
     private oAuthService: OAuthService,
-    public layoutService: LayoutService
+    public layoutService: LayoutService,
+    private router: Router
   ) {
     this.isAuthenticated = oAuthService.hasValidAccessToken();
+    // Listening for navigation events
+    router.events.forEach((event) => {
+      if(event instanceof NavigationEnd) {
+        (window as any).trackingEventsFake('change language:', event);
+      }
+      // NavigationStart
+      // NavigationEnd
+      // NavigationCancel
+      // NavigationError
+      // RoutesRecognized
+    });
   }
 
   logout(): void {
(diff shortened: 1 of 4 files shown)


apps/golden-sample-app/src/app/app.component.ts
```diff
@@ -14,9 +15,21 @@ export class AppComponent {
...
     private oAuthService: OAuthService,
-    public layoutService: LayoutService
+    public layoutService: LayoutService,
+    private router: Router
   ) {
     this.isAuthenticated = oAuthService.hasValidAccessToken();
+    // Listening for navigation events
+    router.events.forEach((event) => {
+      if(event instanceof NavigationEnd) {
+        (window as any).trackingEventsFake('change language:', event);
```
kirillsud, line 25: Why language? Do you mean `location`?

tcorral: You are right, it was just a copy-paste issue.

tcorral: Fixed.

apps/golden-sample-app/src/index.html
```diff
@@ -8,6 +8,33 @@
     <link rel="icon" type="image/x-icon" href="favicon.ico" />
   </head>
   <body>
+    <script type="text/javascript">
+      function trackingEventsFake (eventType, event) {
+        console.log('tracking event type:', eventType, event);
+      }
+
+      // Listening for click events
+      document.addEventListener('click', function (event) {
```
arundathiBB, line 17: @tcorral i like the idea of capturing clicks at the document level. My question though is there are situations when click is not triggered for example while selecting a dropdown option with a keyboard. 

- Do we want to log actions on different set of controls that does not trigger a click (select, input type="texts" and date)

tcorral: What we have been requested and what is more common it's to take care of navigation and click in links and buttons butI do think that what you are asking for will be something to resolve using custom events and maybe at the ui-ang level.

apps/golden-sample-app/src/app/locale-selector/locale-selector.component.ts
```diff
@@ -38,6 +43,8 @@ export class LocaleSelectorComponent implements OnInit {
 
   ngOnInit() {
     this.currentLanguage = this.findLocale(this.localeService.currentLocale);
+    const event = new CustomEvent('data-analytics-change-language', { detail: { locale: this.currentLanguage } });
+    window.dispatchEvent(event);
```
santam85, line 47: This is my favourite way of doing it, perhaps dispatching on document instead of window makes it a little bit easier to inject a value in Angular (I know it's not needed but strict TS validation might throw errors)

apps/golden-sample-app/src/index.html
```diff
@@ -8,6 +8,33 @@
...
+        let target = event.target;
+
+        if (target && target.hasAttribute('data-analytics-id')) {
+          trackingEventsFake('click', event);
+        } else {
+          while(!target.hasAttribute('data-analytics-id') && target.tagName !== 'HTML') {
```
santam85, line 23: clicks within tagged attributes might be tricky, as you might have even multiple ids between the root element and the clicked one, but this is a nice simple way to handle it

libs/transactions/src/lib/views/transactions-view/transactions-view.component.html
```diff
@@ -13,7 +13,7 @@ <h1 *ngIf="title !== ''" data-role="transactions-view__title">
           class="bb-list__item"
           data-role="transactions-view__item-container"
         >
-          <bb-transaction-item [transaction]="item"></bb-transaction-item>
+          <bb-transaction-item [transaction]="item" data-analytics-id="transactions-journey-click-transaction-item"></bb-transaction-item>
```
santam85, line 16: we could use a more meaninful name for the attribute (IE: `data-event-id`)

PR #49, opened 2022-06-23
Example of analytics using events

tcorral:
**DO NOT MERGE, IT'S JUST FOR DEMO PURPOSES!!**

An example of what might be needed to track clicks on transactions and text change in filter input.


kirillsud:
There is a POC that we are discussing. We can close this PR.

PR #54, opened 2022-07-25
Migrate to foundation v7

kirillsud:
To migrate `@backbase/foundation-ang` to v7, dependent Backbase libraries should be migrated. To be able to test the build and library, usages of `positive-pay-journey-ang` were removed, nevertheless, `ui-ang` library is the one that couldn't be excluded from the build.

- [x] wait for `ui-ang` prerelease with removed `foundation-ang` usages
- [x] wait for `ui-ang` 8 public release


package.json
```diff
@@ -35,13 +35,13 @@
...
     "@backbase/positive-pay-check-v1-client-ang": "^1.3.0",
-    "@backbase/positive-pay-journey-ang": "^0.44.0",
-    "@backbase/select-context-widget-ang": "^4.4.0",
-    "@backbase/ui-ang": "^7.5.0",
-    "@backbase/universal-ang": "^7.0.0",
-    "@ng-bootstrap/ng-bootstrap": "^10.0.0",
+    "@backbase/positive-pay-journey-ang": "^2.0.0",
+    "@backbase/select-context-widget-ang": "^5.6.2",
+    "@backbase/ui-ang": "file:../web-ui/dist/libs/ui-ang/backbase-ui-ang-8.0.0-pr.37.tgz",
+    "@backbase/universal-ang": "^8.0.0",
```
santam85, line 43: this should be removed with ui-ang 8 ?

kirillsud: You are right. Removed it. But to make it work with ui-ang 8, I've installed these deps:

- "@ng-select/ng-select": "^8.1.1",
- "bignumber.js": "^9.0.1",
- "ngx-mask": "^13.0.0",

apps/golden-sample-app/src/app/app-routing.module.ts
```diff
@@ -37,17 +37,17 @@ const routes: Routes = [
...
-    loadChildren: () =>
-      import('./ach-positive-pay/ach-positive-pay-journey-bundle.module').then(
-        (m) => m.AchPositivePayJourneyBundleModule
-      ),
-    canActivate: [AuthGuard, EntitlementsGuard],
-    data: {
-      entitlements: triplets.canViewAchRule,
-    },
-  },
+  // {
```
Farzannajipour, line 40: Do we need these lines?

kirillsud: Yes, we will keep it until the journey team will migrate to v7

package.json
```diff
@@ -35,19 +35,21 @@
     "@backbase/arrangement-manager-http-ang": "^1.3.7",
     "@backbase/backbase-theme": "^2.316.0",
```
santam85, line 36: With ui-ang 8 the theme should be imported from ui-ang directly and this dependency removed

.npmrc
```diff
@@ -1 +1,2 @@
-@backbase:registry=https://repo.backbase.com/api/npm/npm-backbase/
+# @backbase:registry=https://repo.backbase.com/api/npm/npm-backbase/
+@backbase:registry=https://artifacts.backbase.com/api/npm/npm-backbase/
```
igord, line 2: Can we merge this before v7 is on repo?

apps/golden-sample-app/src/app/user-context/user-context.guard.ts
```diff
@@ -0,0 +1,71 @@
...
+    console.log('cookieIsSet', cookieIsSet);
+    if (!cookieIsSet) {
+      return this.getSelectContextUrlTree();
+    }
+
+    return (
+      this.cookieValid ||
+      this.validateUserContextCookie().pipe(map(isValid => isValid || this.getSelectContextUrlTree()))
+    );
+  }
```
igord, line 49: Maybe explain the logic in comment, it is difficult to understand why and what are you doing here.

.npmrc
santam85: an RC of ui-ang v8 was just promoted to repo, AFAIK that is the only reason why we needed artifacts here...

package.json
kirillsud: I can't see this package in dependencies for `ui-ang`. Is it a task to do?

santam85: they basically embedded the theme in ui-ang, so the backbase-theme package will not be published anymore from ui-ang 8 onwards, but the sass files will be part of ui-ang itself

.npmrc
kirillsud: They promoted it, so I reverted this

package.json
kirillsud: After removing theme package and applying suggested changes I received a broken layout:

image

image

I've fixed it by overriding `z-index` for the `bb-layout__topbar`: https://github.com/Backbase/golden-sample-app/pull/54/commits/fecd0f9714588f1eeae23e597d0953b5f5c8ab43#diff-c0301706b3c2178287213cb76542bff338c966bf41db75421789a9be0efc58f9R16

```
.bb-layout .bb-layout__topbar {
  z-index: $zindex-modal-backdrop + 1;
}
```

kirillsud: FYI @okadurin

santam85: Odd I didn't have to do that in my branch 🤷

apps/golden-sample-app/src/app/user-context/user-context.guard.ts
kirillsud: I've rewritten this guard a bit. @igord could you please review it again?

santam85 (approved):

PR #57, opened 2022-08-15
WF-513  Gha pipeline

manisha-backbase:


Farzannajipour (approved):

kirillsud (changes requested):

kirillsud:
I couldn't find an e2e test run for the workflow. What is the reason that you didn't add it?

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Derive appropriate SHAs for base and head for `nx affected` commands
+        uses: nrwl/nx-set-shas@v2
+        with:
+          main-branch-name: ${{ github.base_ref }}
+
+      - name: Setup Node with version ${{ matrix.node-version }}
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+      - run: curl -s -u"${{ secrets.MAVEN_USERNAME }}:${{ secrets.MAVEN_PASSWORD }}" https://repo.backbase.com/api/npm/npm-backbase/auth/backbase > .npmrc
```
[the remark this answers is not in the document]
kirillsud: Also, we shouldn’t override the original `.npmrc` file. We can create a custom `.npmrc` (e.g. `.npmrc-user`) file and specify it as a user config file:

```
curl -s -u"${{ secrets.MAVEN_USERNAME }}:${{ secrets.MAVEN_PASSWORD }}" https://artifacts.backbase.com/api/npm/auth | sed "s/^/\/\/artifacts.backbase.com\/api\/npm\/npm-backbase\/:/" >> .npmrc-user
curl -s -u"${{ secrets.MAVEN_USERNAME }}:${{ secrets.MAVEN_PASSWORD }}" https://repo.backbase.com/api/npm/auth | sed "s/^/\/\/repo.backbase.com\/api\/npm\/npm-backbase\/:/" >> .npmrc-user
export NPM_CONFIG_USERCONFIG=.npmrc-user
```

In this way, we keep project's npmrc and add tokens for repo and artifacts registries.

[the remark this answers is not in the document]
manisha-backbase: Hi Kirill, But what is the issue in creating .npmrc in a container instance? Once the job completes the container will be deleted. As this new created file is not pushed to the repo, I feel this should not be an issue.

FYI: I am not in favour of using curl at all. This seems like a hack. setup-node action already has this functionality where it create a npmrc file for the private registry but needs a auth-token. In our case for repo we don't have such thing.
https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages#example-workflow

manisha-backbase:
> 

Hi Kirill,

I just added the jobs mentioned in the ticket. But I will add the job  for e2e.

manisha-backbase:
> I couldn't find an e2e test run for the workflow. What is the reason that you didn't add it?

@kirillsud  Does e2e test pass currently? Could you please confirm.

.github/workflows/pull-request.yml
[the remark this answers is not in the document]
santam85: I can add the auth token to the secrets, no problem
We need one for repo/artifactory or the public npmjs.org?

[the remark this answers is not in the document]
kirillsud: @santam85 only for `repo`. We can't download from `artifacts` without VPN.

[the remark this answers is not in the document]
kirillsud: @manisha-backbase we have to keep an original `.npmrc` because it is a part of the source code. We can go with your solution but by appending to the existing `.npmrc` (with `>>` instead `>`)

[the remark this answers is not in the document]
manisha-backbase: @kirillsud  I have made the changes

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,55 @@
...
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [14.x]
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
```
santam85, line 22: Isn't this the default?

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,55 @@
...
+          fetch-depth: 0
```
santam85, line 23: Is this needed for nx?

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,55 @@
...
+  pull_request:
+    branches:
+      - main
+  
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
```
santam85, line 16: doesn't make much sense using a matrix strategy if we are using only one variant. I recommend updating this to node 16 and just use that in the setup-node step

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,55 @@
...
+        run: npm run format:check  
+
+      - name: Unit test
+        run: npm run affected:test
+
+      - name: Check Linting
+        run: npm run affected:lint 
+      
+      - name: Install Playwright
+        run: npx playwright install --with-deps
```
santam85, line 52: I believe this should be `npx playwright install-deps`, as the config files should already be generated

kirillsud: As I understand, `playwright install-deps` installs only system dependencies for browsers, but `playwright install` will install supported browsers. There is a playwright container that can be used to run workflow steps - https://playwright.dev/docs/ci#github-actions-via-containers. With it, we can skip this step.

Also, this step takes about 1 minute, which is also a good argument to get rid of it.

image

The only thing I'm not sure about is which node version is installed in the playwright container.

.github/workflows/pull-request.yml
kirillsud: I guess yes, because we use `nx affected`

.github/workflows/pull-request.yml
manisha-backbase: @santam85  I removed it for now.

.github/workflows/pull-request.yml
santam85: yeah that's what I was afraid of. We can optimize later, but this is a good low-hanging fruit...

.github/workflows/pull-request.yml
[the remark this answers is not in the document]
santam85: Token requested with https://backbase.atlassian.net/servicedesk/customer/portal/19/SUPPORT-38709

.github/workflows/pull-request.yml
manisha-backbase: @santam85  As this is triggered on pull request I am explicitly setting it to head reference of the PR to get till the last commit of the feature branch.
I am not sure what will it default to?

santam85: actually it's not the default: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

Basically the default is the merge commit of the PR, which is a better choice imho than the PR head (as that might not contain commits that are present on the target branch, failing to test the resulting merge and using the PR source branch instead)

.github/actions/setup-node/action.yml
```diff
@@ -0,0 +1,64 @@
+name: "Setup Node Environment"
```
santam85, line 1: I don't think this whole workflow is actually needed? the default `setup-node` action already manages a cache for the node_modules folder since v3 (as described at https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) 

So I think we can greatly simplify the workflow?

manisha-backbase: Oh I missed this that in V3 it caches the node_modules.

package.json
```diff
@@ -12,6 +12,9 @@
     "build": "nx build --configuration production",
     "build:stats": "ng build --configuration production --stats-json",
     "build:docker": "docker build -t golden-sample-app:latest .",
+    "affected:build": "nx affected:build --base=origin/${CHANGE_TARGET:-main} --configuration=production",
+    "affected:lint": "nx affected:lint --base=origin/${CHANGE_TARGET:-main}",
+    "affected:test": "nx affected:test --base=origin/${CHANGE_TARGET:-main}",
```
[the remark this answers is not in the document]
manisha-backbase: I think there is no harm it putting it in package.json so that It can be reused.

.github/workflows/pull-request.yml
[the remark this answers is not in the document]
manisha-backbase: We can change the implementation later on. Once, we get some information to use the auth token for the repo.
I will resolve this conversation for now.

.github/workflows/pull-request.yml
manisha-backbase: I have added the e2e tests in a separate workflow and it works fine in parallel. We can optimise this late on as well.

kirillsud (approved):

PR #68, opened 2022-09-28
 Test against all supported Angular versions.

Farzannajipour:


.github/workflows/e2e.yml
```diff
@@ -7,7 +7,11 @@ on:
 jobs:
   playwright:
     name: 'Playwright Tests'
+    timeout-minutes: 60
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [14.x, 16.x]
```
santam85, line 14: Why test multiple node versions? Let's stick to LTS and save GHA minutes, it's not a node serverside app...

.github/workflows/e2e.yml
```diff
@@ -28,4 +32,13 @@ jobs:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
 
       - name: Run Playwright tests
-        run: npm run e2e
```
santam85, line 31: I'd rather us using a builder for e2e tests here. I believe we had an example configured that used a generic builder capable of supporting playwright tests? I really don't want to run the app under test using the playwright config as a reference best practice we give to customers and teams

.github/workflows/e2e.yml
```diff
@@ -7,7 +7,11 @@ on:
 jobs:
   playwright:
     name: 'Playwright Tests'
+    timeout-minutes: 60
```
santam85, line 10: This is way to high: a mistake here will cost us 1h of GHA minutes. e2e tests for an app as small as this one should finish in less than 5-10, including the build.

.github/workflows/e2e.yml
santam85: This is a good example: https://dev.to/darlanalves/playwright-angular-cli-39ab

.github/workflows/pull-request.yml
```diff
@@ -25,11 +42,16 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: 16
+          node-version: ${{ matrix.node-version }}
           cache: 'npm'
           registry-url: 'https://repo.backbase.com/artifactory/api/npm/npm-backbase/'
           scope: '@backbase'
 
+      - name: Use Angular ${{ matrix.angular-version }}
```
kirillsud, line 50: I'm not sure that we can apply different Angular versions for the same code source. For example, if we upgrade the `main` branch to Angular v14 and start using standalone components, the app couldn't be built with a lower version.

My suggestion is to use a matrix parameter for repo branch names and specify a branch name in `actions/checkout` step. We should prepare a branch for each supported Angular version.

kirillsud (changes requested):
I think it better to prepare a separate workflow for the testing against all supported Angular versions which will be triggered manually or by schedule. But definitely, we do need to run it for each PR commit.

.github/workflows/pull-request.yml
Farzannajipour: That's a valid point. we don't need to run it for every pull request. If the event occurred on a particular repository branch, the workflow files should be in that branch's repository. I created the branches and split the workflows.

PR #71, opened 2022-09-30
Fix for opinionated auth guard and interceptor

sfuenffinger:
Kirill spotted a problem with the select-context route and logging out. These 2 issues should be solved by adding the AuthService to the app initializer.


apps/golden-sample-app/src/app/app.module.ts
```diff
@@ -80,7 +80,12 @@ import { IdentityAuthModule } from '@backbase/identity-auth';
     {
       provide: APP_INITIALIZER,
       multi: true,
-      deps: [OAuthService, CookieService, AuthEventsHandlerService],
+      deps: [
+        OAuthService,
```
santam85, line 84: this doesn't look right, it doesn't matches the parameters of the factory

apps/golden-sample-app/src/app/app.module.ts
```diff
@@ -80,7 +80,12 @@ import { IdentityAuthModule } from '@backbase/identity-auth';
...
+        AuthService,
+        CookieService,
+        AuthEventsHandlerService,
```
santam85, line 87: this doesn't look right, it doesn't matches the parameters of the factory

```suggestion
        AuthService,
        CookieService
```

apps/golden-sample-app/src/app/app.module.ts
```diff
@@ -10,6 +10,7 @@ import {
   EntitlementsModule,
   ENTITLEMENTS_CONFIG,
 } from '@backbase/foundation-ang/entitlements';
+import { AuthService, IdentityAuthModule } from '@backbase/identity-auth';
```
santam85, line 13: What is the difference between this `AuthService` and `angular-oauth2-oidc` `OAuthService`?

sfuenffinger: AuthService exposes an `isAuthenticated` observable which is used in the AuthGuard. The OidcAuthService handles refresh, logout, creating login urls etc.

PR #74, opened 2022-10-06
Additional example for observable route params

kirillsud:
I've added an additional example that demonstrates re-using of the same component when changing only route parameters in the navigation:

image


libs/transactions/src/lib/views/transaction-details-view/transaction-details-view.component.scss
```diff
@@ -0,0 +1,4 @@
+.header {
```
Farzannajipour, line 1: I would recommend using BB design system classes or in this case Bootstrap classes instead of creating the custom style.

santam85 (approved):

PR #79, opened 2022-10-07
Add foundation-ang manual trigger (and push workflow) for test the different versions

Farzannajipour:


.github/workflows/foundation-ang-trigger.yml
```diff
@@ -0,0 +1,47 @@
+# Controls when the action will run. The workflow runs when manually triggered using the UI or API
+
+name: Foundation_Ang_On_Demand
+
+on:
+  workflow_dispatch:
```
manisha-backbase, line 6: Can we have one single file for on dispatch and on push ?

.github/workflows/push-foundation-ang-checker.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      - name: Install Dependencies
+        run: npm ci --legacy-peer-deps
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}          
+
+      # Check with the current foundation-ang version
+      - if: ${{ env.FOUNDATION_ANG_VERSION }} != npm view @backbase/foundation-ang version
+        run: |
+          npm run affected:build
```
manisha-backbase, line 42: Why are we using affected?

.github/workflows/foundation-ang-trigger.yml
Farzannajipour: I added the push event to check the workflow, it will be removed after testing the manual foundation and trigger on the master.

.github/workflows/foundation-ang-trigger.yml
```diff
@@ -0,0 +1,47 @@
...
+          registry-url: 'https://repo.backbase.com/artifactory/api/npm/npm-backbase/'
+          scope: '@backbase'
+
+      - name: Install Dependencies
+        run: npm ci --legacy-peer-deps
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}   
+
+      # Check with the current foundation-ang version
+      - if: github.event.inputs.foundation_ang_version != npm view @backbase/foundation-ang version
```
kirillsud, line 42: What is this check for?

.github/workflows/foundation-ang-trigger.yml
```diff
@@ -0,0 +1,47 @@
+# Controls when the action will run. The workflow runs when manually triggered using the UI or API
+
+name: Foundation_Ang_On_Demand
+
+on:
+  workflow_dispatch:
+    # Inputs the workflow accepts.
+    inputs:
+      foundation_ang_version:
```
kirillsud, line 9: Am I missing something? I remember we decided to install a specified version of `foundation-ang`. I can't see this below.

.github/workflows/foundation-ang-trigger.yml
kirillsud: Why don't add another trigger instead of adding another workflow?

kirillsud (changes requested):

.github/workflows/foundation-ang-trigger.yml
Farzannajipour: Done

.github/workflows/push-foundation-ang-checker.yml
Farzannajipour: We don't need to use `affected:lint` and `format:check` for the manual trigger, there won't be a change by having a new version of foundation-ang. Although to figure out the changes in the workspace for the build configuration that can be affected by this change, I think it's required.

kirillsud: The point from Manisha is about using build all instead of affected because nothing is affected in manual workflow execution

.github/workflows/foundation-ang-trigger.yml
```diff
@@ -0,0 +1,56 @@
...
+          cache: 'npm'
+          registry-url: 'https://repo.backbase.com/artifactory/api/npm/npm-backbase/'
+          scope: '@backbase'
+
+      - name: Install Dependencies
+        run: npm ci --legacy-peer-deps
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+        if: ${{ github.event_name == push && github.ref_name == dev/angular13 }}
```
kirillsud, line 44: Why do we need this condition?

kirillsud (changes requested):

.github/workflows/push-foundation-ang-checker.yml
Farzannajipour: I changed it.

.github/workflows/foundation-ang-trigger.yml
Farzannajipour: We have two separate events in the `foundation-ang-trigger.yml`:

- push
- workflow_dispatch

In both events, I overwrote the foundation-ang version.  The first event uses the environment variable and the second event uses the user's input which is the same as the environment variable (7.0.0) but can be different based on the user input. So, based on the event name, I have the different conditions to overwrite foundation-ang suitable version.

.github/workflows/foundation-ang-trigger.yml
```diff
@@ -0,0 +1,56 @@
+# Controls when the action will run. The workflow runs when manually triggered using the UI or API
+
+name: Foundation_Ang_On_Demand
+
+on:
+  push:
+    branches:
+      - dev/angular13
```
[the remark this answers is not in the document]
Farzannajipour: I don't think we need to check every push event for the main branch. We have two branches (dev/angular12, dev/angular13) which are testable with push event. Next to that, by manual trigger, we can try it on every branches and also every foundation-ang we want.

.github/workflows/foundation-ang-trigger.yml
```diff
@@ -0,0 +1,53 @@
...
+      - dev/angular12
+  workflow_dispatch:
+    # Inputs the workflow accepts.
+    inputs:
+      foundation_ang_version:
+        description: 'Foundation-ang version'
+        required: true
+        default: '7.0.0'
```
igord, line 16: How can parameter be both required and have default value?

.github/workflows/foundation-ang-trigger.yml
```diff
@@ -0,0 +1,53 @@
+# Controls when the action will run. The workflow runs when manually triggered using the UI or API
+
+name: Foundation_Ang_On_Demand
+
+on:
+  push:
+    branches:
+      - dev/angular13
+      - dev/angular12
```
igord, line 9: Shouldn't those branches be part of the PR as well?

.github/workflows/foundation-ang-trigger.yml
Farzannajipour: The workflow_dispatch properties can have different types like string or boolean.
To run the workflow, it needs to have foundation_ang_version as a string (7.0.0 in this case). So if it's empty it's informed and doesn't allow running workflow without it. It means it's mandatory to have this value to run.

.github/workflows/foundation-ang-trigger.yml
Farzannajipour: Yes, They are already in place and sync with the Angular versions.

manisha-backbase (approved):

.github/workflows/foundation-ang-trigger.yml
kirillsud: As I understand the `default` value will be put into the form when you trigger the workflow (e.g **Log level**):

image

But I agree that we can remove it because it is always required to enter by the developer.

.github/workflows/foundation-ang-trigger.yml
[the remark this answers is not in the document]
kirillsud: The idea for the `push` trigger was to test the pipeline before merging the PR to `main`. So, I'd like to see on [this page](https://github.com/Backbase/golden-sample-app/pull/79/checks) that your pipeline works. That's it. I don't want to see it later AFTER merging the PR.

Please, change it to:

```
on:
  pull_request:
    branches:
      - main
```

check if the workflow works and then remove it from the PR.

.github/workflows/foundation-ang-trigger.yml
Farzannajipour: That works similarly to the forms input. So, because it is part of the workflow to overwrite the foundation-ang version, it should be mandatory so we need to keep it.

igord: ok thanks.
I would suggest then to remove both required and default.
So if it is not set, you `npm info` to get the latest version and use it

.github/workflows/foundation-ang-trigger.yml
igord: I am a bit lost here... So they were already created in some other PR?
Is there any other reason to have them apart from testing different versions?

.github/workflows/foundation-ang-trigger.yml
Farzannajipour: what are the benefit of not having required and default

.github/workflows/foundation-ang-trigger.yml
Farzannajipour: There are two points: 

1. The reason for using branches (not tags) is by default they are available as one of the options to use the workflow. Next to that, we also have the foundation-ang version. So, the testing would be straightforward to use the pair of branches and the foundation-ang version. 
image

2. I created those two branches manually. I agree with you that we can create them automated (based on git tag latest commits that use those angular versions) but we should take into account that every six months we need to have a new branch.

.github/workflows/foundation-ang-trigger.yml
Farzannajipour: Most of the time, I would consider the first and second options, although, the third option is more or less the same as the second. The use case of it is having a bug from/in a specific version of the libraries and it gives us the capability to debug it better. 
On top of that, based on [github-workflow.json](url) , we can't remove `required`, It is mandatory to have it. I will turn it to false and I will remove the default value.

.github/workflows/foundation-ang-trigger.yml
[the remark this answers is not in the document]
Farzannajipour: https://github.com/Backbase/golden-sample-app/actions/runs/3236043470

.github/workflows/foundation-ang-trigger.yml
```diff
@@ -0,0 +1,60 @@
...
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name:  Install Foundation-ang if foundation_ang_version's input is not empty
+        if: "${{ github.event.inputs.foundation_ang_version != '' }}"
+        run: npm install @backbase/foundation-ang@${{ github.event.inputs.foundation_ang_version }}
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}        
+
+      - name: Install Foundation-ang if foundation_ang_version's input if empty
+        if: "${{ github.event.inputs.level == '' }}"
```
igord, line 49: `github.event.inputs.level` or `github.event.inputs.foundation_ang_version`?

igord (commented):
Do I understand it right that there is no input for branches to test, which makes it possible to run this on any git ref?

Farzannajipour:
> Do I understand it right that there is no input for branches to test, which makes it possible to run this on any git ref?

Yes indeed, it's possible to test by the different branches or tags.

PR #81, opened 2022-10-10
Update package.json

danbilokha:
Added hint for engines since running `npm ci` fails for node16 (tested 16.17.1), node 14.0.0


kirillsud (changes requested):
We are using node 16 to build the app. We have GitHub workflows that are using node 16 also. So I think the root cause of your issue in your local environment. In any case - specifying an engine version is a good idea. But, I believe, it should be v16

PR #83, opened 2022-10-13
WF-830: test foundation-ang with all supported angular versions

kirillsud:
The workflow job uses matrix parameters to checkout and test different git references that are linked with different Angular versions. Example https://github.com/Backbase/golden-sample-app/actions/runs/3246197479:

image

There are two tags for two Angular versions: `angular@v12` and `angular@v13`. The workflow `tag-angular-version` is used to update these tags.


manisha-backbase (approved):

Farzannajipour (commented):
Looks good to me.

PR #95, opened 2022-10-31
fix: Fixed release drafter template

manisha-backbase:


.github/workflows/release-drafter.yml
```diff
@@ -38,4 +43,22 @@ jobs:
           with:
             config-name: release-drafter-transfer.yml
           env:
-            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
\ No newline at end of file
+            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  tag:
+    runs-on: ubuntu-latest
```
kirillsud, line 49: Can we add a more descriptive name for the job?

.github/workflows/release-drafter.yml
```diff
@@ -4,17 +4,21 @@ on:
   push:
     branches:
       - main
+      - test/angular-v*
```
kirillsud, line 7: We can remove this line because in the ideal situation we don't need these branches. And if we create them, we will modify the condition for the workflow in a branch.

kirillsud (approved):

kirillsud (approved):

PR #114, opened 2022-11-21
Remove API proxy for local server [RFF-2397]

kirillsud:
After introducing `X-User-Context` it became possible to use the API server with a different origin. It allows:

1. Use `nx serve` without the API proxy
2. Solve an issue with the application's local docker image

The solution was taken from the Employee Web App implementation.

Additionally, useless environment variables have been removed from the `docker-compose.yml`.


santam85 (approved):
Yes, perfect, thanks @kirillsud !
One small disclaimer: the configuration without the proxy works for everything except file downloads. The teams that own capabilities with file download features are busy removing this limitation, and the work should be completed by end of year.

PR #116, opened 2022-12-16
Add activity monitor component example - IDWOJ-1438

vinylhero:


santam85:
Looks like credentials do not work anymore, investigating...

package.json
```diff
@@ -59,6 +60,7 @@
     "quill": "^1.3.7",
     "rxjs": "^7.5.6",
     "tslib": "^2.3.0",
+    "webpack": "^5.75.0",
```
kirillsud, line 63: @vinylhero why `webpack` was added to dependencies?

PR #127, opened 2023-02-03
migrate to web-base v2 and fix locale sync with identity

kirillsud:
The PR includes migration from `web-base` v1 to `web-base` v2 and fixes for locale sync between Identity and web app.

After the migration to `web-base` v2 I've found that the locale selector component has been broken. The reason is a fix in `web-base`:

> A bug was found for multi-locale apps when the root request made an internal redirect (without a URL update) to a specific locale. This behavior led to the wrong Angular route resolution. For example, the root request /golden-sample made an internal redirect to the /golden-sample/en/index.html file, and Angular resolved the path as a sub-route. As a result, the user received an updated wrong path /golden-sample/en/golden-sample. This error is invisible if the application redirects from any unknown route to the landing page. This was fixed by making the permanent redirect to a specific locale if the corresponding locale exists in the application.

The redirect issue was fixed in `LocalesService`:

https://github.com/Backbase/golden-sample-app/pull/127/files#diff-3596693bb80ae5016e9be97785fee4f19b53386a62a97e6c70c438693e8c6f94R20-R36

Also, the PR includes a proposal for [WF-1019](https://backbase.atlassian.net/browse/WF-1019) to make a switch to the correct locale in the application when the user has logged in Identity. @vinylhero @sfuenffinger please review my changes for `AuthEventsHandlerService`:

https://github.com/Backbase/golden-sample-app/pull/127/files#diff-b7b19f9acee048d10db00854b81ec439e5c2dc4570ac457da1364c59a97e0ab6R62-R64

### DIFF
--- a/apps/golden-sample-app/src/app/locale-selector/locales.service.ts
+++ b/apps/golden-sample-app/src/app/locale-selector/locales.service.ts
@@ -21,16 +21,24 @@ export class LocalesService {
     const currentLocale = this.locale;
 
     // Get base href without locale
-    const baseHref = this.location.getBaseHref().replace(new RegExp(`/${currentLocale}/?$`), '');
+    const baseHref = this.location
+      .getBaseHref()
+      .replace(new RegExp(`/${currentLocale}/?$`), '');
 
-    const cookieValue = `${encodeURIComponent('bb-locale')}=${encodeURIComponent(locale)}`;
+    const cookieValue = `${encodeURIComponent(
+      'bb-locale'
+    )}=${encodeURIComponent(locale)}`;
     const cookiePath = `path=${baseHref === '' ? '/' : baseHref}`;
 
-    this.document.cookie = [cookieValue, cookiePath, COOKIE_ATTRIBUTES].join('; ');
+    this.document.cookie = [cookieValue, cookiePath, COOKIE_ATTRIBUTES].join(
+      '; '
+    );
 
     if (locale !== currentLocale) {
       // Get path without base href and locale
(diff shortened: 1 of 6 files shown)


apps/golden-sample-app/src/app/locale-selector/locales.service.ts
```diff
@@ -1,22 +1,39 @@
...
   ) {}
-  setLocaleCookie(value: string) {
-    this.document.cookie =
-      encodeURIComponent('bb-locale') +
-      '=' +
-      encodeURIComponent(value) +
-      `;${COOKIE_ATTRIBUTES}`;
+
+  setLocale(locale: string) {
+    const currentLocale = this.locale;
```
manisha-backbase, line 21: Why are we using a const and not adding it in the constructor itself?

apps/golden-sample-app/src/app/locale-selector/locales.service.ts
```diff
@@ -1,22 +1,39 @@
...
+
+    // Get base href without locale
+    const baseHref = this.location.getBaseHref().replace(new RegExp(`/${currentLocale}/?$`), '');
+
+    const cookieValue = `${encodeURIComponent('bb-locale')}=${encodeURIComponent(locale)}`;
+    const cookiePath = `path=${baseHref === '' ? '/' : baseHref}`;
+
+    this.document.cookie = [cookieValue, cookiePath, COOKIE_ATTRIBUTES].join('; ');
```
manisha-backbase, line 29: Do we need to set this cookie if locale === currentLocale

PR #129, opened 2023-03-14
WF-1072-test-the-new-theme-changes-in-ui-ang-9-for-gsa

manisha-backbase:


package.json
```diff
@@ -28,12 +28,12 @@
     "@angular/cdk": "^14.2.7",
     "@angular/common": "14.2.12",
     "@angular/compiler": "14.2.12",
-    "@angular/core": "^14.2.12",
+    "@angular/core": "^14.3.0",
```
ankypant, line 31: Question:
Why were `core` & `router` updated to `14.3.0`?

ankypant (approved):

package.json
manisha-backbase: @ankypant This is a minor version and its okii to put it to next minor as there is already a caret and also the migration guide for the ui-ang 9 had this angular version.
https://backbase.io/developers/angular/design-system/ui-ang/9.0.0/additional-documentation/migration-notes.html

kirillsud: Agree with a caret, but it should be in place for all angular packages then. Currently there is an issue with `peerDependencies`:

```
npm view @angular/router@14.3.0 peerDependencies
{
  '@angular/core': '14.3.0',
  '@angular/common': '14.3.0',
  '@angular/platform-browser': '14.3.0',
  rxjs: '^6.5.3 || ^7.4.0'
}
```

As you can see, `@angular/router` depends on `@angular/common` 14.3.0, but in the `package.json` we have 14.2.12 without a caret.

Dockerfile
```diff
@@ -1,3 +1,3 @@
-FROM repo.backbase.com/backbase-docker-releases/web-base:2.0.1
+FROM repo.backbase.com/backbase-docker-releases/web-base:latest
```
kirillsud, line 1: We have a requirement from the security team not to use the latest tag because it is dynamic and leads to uncontrollable updates for the app.

kirillsud (changes requested):

santam85:
@manisha-backbase what are we doing with this PR? Is it still needed?

manisha-backbase:
> @manisha-backbase what are we doing with this PR? Is it still needed?

This is for updating to ui-ang 9. There were some UI issues with the latest version of ui-ang 9. Hence, this has been put on hold till the bug is fixed.

santam85:
@manisha-backbase can you please rebase?

PR #165, opened 2023-05-11
https://backbase.atlassian.net/browse/WF-1148: Update structure of transfer Journey

manisha-backbase:
Hi All,
As part of the "create a journey" web devkit, we have restructured the folder structure of Journeys for golden sample app as recommended by the journey [cook book](https://backbase.atlassian.net/wiki/spaces/GUIL/pages/3939205398/Internal+structure). While doing this, I had the following doubts:

1. I had a file defining the constants and  models which has to be shared in ui lib, feature lib and data-access lib. To handle this case, I created a new shared lib inside the journey  `libs/transfer/shared-data`. The module boundary for this that all the other libraries can import from this. But this library can only import from itself. 
2. Currently in many Journeys , there is a view folder inside shell library of a journey. The confusion here is what should go in views folder in shell type lib and feature type lib. Should all the routable smart components go in feature lib? For now, I added all the smart routable components in a feature lib.

I would really appreciate your thoughts on the above two points. As we will be adding the references of GSA in our devkit guides. So, these two points are  important to address, so that we define a flexible yet a  standard  for journey structure. 

@santam85 @cheshirsky @jordydejong @ankypant @tcorral


libs/transfer-journey/shared-data/src/lib/config/model/Account.ts
```diff
@@ -0,0 +1,11 @@
+export interface Account {
```
ankypant, line 1: Let's rename this to account.ts with a small `a`

libs/transfer-journey/shared-data/project.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "transfer-journey-shared-data",
+  "$schema": "../../../node_modules/nx/schemas/project-schema.json",
+  "projectType": "library",
+  "sourceRoot": "libs/transfer-journey/shared-data/src",
```
ankypant, line 5: I thought that we are going to have a separate `shared` lib.
Is this just for the `transfer-journey` or is it being shared with other journeys as well?

libs/transfer-journey/shell/src/lib/make-transfer-journey-store-guard.ts
```diff
@@ -0,0 +1,17 @@
...
+  canActivate() {
+    const state = this.router.getCurrentNavigation()?.extras?.state;
+    if (!state || !state['transfer']) {
+      return this.router.parseUrl('../make-transfer');
+    }
+    return true;
+  }
+  constructor(
+    private readonly router: Router,
+    private readonly route: ActivatedRoute
```
ankypant, line 15: We are not using `route` anywhere so this could be removed.
Does it make sense to do this in this PR, it's just for updating journeys to follow DDD?

libs/transfer-journey/shell/src/index.ts
```diff
@@ -0,0 +1,7 @@
+export * from './lib/transfer-journey-shell.module';
```
ankypant, line 1: Nice work on cleaning up the `shell` lib 🚀

libs/transfer-journey/shared-data/project.json
manisha-backbase: This data is not relevant to any other Journey. So why do we need to create a shared folder for other Journeys.

libs/transfer-journey/shell/src/lib/make-transfer-journey-store-guard.ts
manisha-backbase: Lets not review the code for this PR . Though, I will remove this for now.

ankypant (approved):

PR #166, opened 2023-05-11
[https://backbase.atlassian.net/browse/WF-1148] Fix/update transactions journey structure

manisha-backbase:
Added DDD structure for transactions Journey


libs/transactions-journey/data-access/src/lib/services/arrangements.service.spec.ts
ankypant, line 1: It would be nice if we can group the service & the spec file in a folder for an example 
arrangements folder would contain arrangement.service and the spec file.

ankypant (approved):

PR #176, opened 2023-05-16
test: node_modules caching

ankypant:


manisha-backbase (approved):

.github/actions/node-step/action.yml
```diff
@@ -0,0 +1,56 @@
+name: "Setup Node Environment"
+description: >
+  Set up node environment and install NPM dependencies.
+  Node modules and Angular cache are cached, and installation is skipped if retrieved.
+inputs:
+  working-directory:
```
manisha-backbase, line 6: Do we need this input in our case?

ankypant: We could omit it as well and hardcode it in the script, but I wanted to keep this as well 
since it's optional and doesn't needs to be specified as input for this action.

PR #233, opened 2024-05-14
Feature/gsa with api sandbox

ankypant:


tcorral:
This PR is not needed anymore because it has been addressed in a different PR that has been already merged.

PR #242, opened 2024-08-22
Journeys in environment configuration

arrilight:
This PR proposes a file-structure change for the apps: journey bundles will now live in shared libraries instead of inside of the app. This way we can dynamically change journey-list based on the configuration of the environment, having multiple apps with different journeys inside of one.


libs/journey-bundles/src/ach-positive-pay/bundle.ts
unlimitedjest, line 1: The file name `bundle` is a bit confusing, especially with the more specific file name including `bundle`.  Maybe including `route` in the name somewhere would help clarify at a glance?

libs/journey-bundles/src/transactions/transactions-journey-bundle.module.ts
unlimitedjest, line 1: Picking transactions as an example, but my comment would apply in general:

It could be useful to separate customer defined journey configuration options from options set in the product.

For example, I'd have a file `journey-config-overrides` that exports
```
export const journeyConfigOverrides: Partial<TransactionsJourneyConfiguration> = {
   showPendingOnTop: true,
   showCheckImages: true,
   accountInfoProperties: { ...goes here }
}
```

that would then be consumed by the generated journey bundle file here:
```
{
      provide: TransactionsJourneyConfiguration,
      deps: [EnvironmentService],
      useFactory: (envService: EnvironmentService) =>
        ({
          pageSize: 10,
          slimMode: envService.environment?.common.designSlimMode,
          ...journeyConfigOverrides
        } as TransactionsJourneyConfiguration),
}
```
Basically, during an upgrade, having this separation would 1) ensure we preserve journey configuration options set during development and 2) make it more transparent which values were set OOTB and which were added by the customer.

unlimitedjest (commented):
All in all I think this is a solid approach, and I'm excited for the changes!

libs/journey-bundles/src/ach-positive-pay/navigation.ts
```diff
@@ -0,0 +1,8 @@
+import { Navigation, triplets } from '@backbase-gsa/shared';
+
+export const ACH_POSITIVE_PAY_JOURNEY_NAVIGATION: Navigation = {
+  name: $localize`Ach Positive pay link@@main.ach-positive-pay.link.text:ACH Positive Pay`,
```
daiscog, line 4:
```suggestion
  name: $localize`:Ach Positive pay link@@main.ach-positive-pay.link.text:ACH Positive Pay`,
```

libs/journey-bundles/src/custom-payment/navigation.ts
```diff
@@ -0,0 +1,7 @@
+import { Navigation } from '@backbase-gsa/shared';
+
+export const CUSTOM_PAYMENT_JOURNEY_NAVIGATION: Navigation = {
+  name: $localize`Make a Payment Link@@main.make-a-payment.link.text:Make a payment`,
```
daiscog, line 4:
```suggestion
  name: $localize`:Make a Payment Link@@main.make-a-payment.link.text:Make a payment`,
```

libs/journey-bundles/src/transactions/navigation.ts
```diff
@@ -0,0 +1,8 @@
+import { triplets, Navigation } from '@backbase-gsa/shared';
+
+export const TRANSACTIONS_JOURNEY_NAVIGATION: Navigation = {
+  name: $localize`transactions list link@@main.transactions.link.text:Transactions List`,
```
daiscog, line 4:
```suggestion
  name: $localize`:transactions list link@@main.transactions.link.text:Transactions List`,
```

libs/journey-bundles/src/transactions/transactions-journey-bundle.module.ts
```diff
@@ -23,10 +25,12 @@ import { CommonModule } from '@angular/common';
...
-      useValue: {
-        pageSize: 10,
-        slimMode: environment.common.designSlimMode,
-      } as TransactionsJourneyConfiguration,
+      deps: [EnvironmentService],
+      useFactory: (envService: EnvironmentService) =>
+        ({
+          pageSize: 10,
+          slimMode: envService.environment?.common.designSlimMode,
+        } as TransactionsJourneyConfiguration),
```
daiscog, line 33: I like to avoid casting wherever possible as it can mask some underlying type errors.  Here, for example, `slimMode` is defined as just `boolean` in the `TransactionsJourneyConfiguration` type but the object being cast as that type could actually have `boolean | undefined` as the type of `slimMode`.

Refactoring to have an explicit return type on the factory exposes this as a compile error:

```suggestion
      useFactory: (envService: EnvironmentService): TransactionsJourneyConfiguration =>
        ({
          pageSize: 10,
          slimMode: envService.environment?.common.designSlimMode,
        }),
```

daiscog: (I appreciate this isn't part of the point of the PR - there's a lot of stuff in the GSA that could do with cleaning up.)

apps/golden-sample-app/src/app/app.component.html
```diff
@@ -47,128 +47,52 @@
...
-            class="bb-layout__horizontal-nav-item-description"
-            i18n="Ach Positive pay link@@main.ach-positive-pay.link.text"
-          >
-            ACH Positive Pay
+          <span class="bb-layout__horizontal-nav-item-description">
+            {{ journey.name }}
           </span>
         </a>
       </li>
+      } @else {
```
daiscog, line 74: There's a lot of duplication of html structure here, just because `*bbIfEntitlements` will display nothing when given undefined.  Maybe we can avoid this using an ng-template...

```html
      @for (journey of journeys; track journey) {

        <ng-template #journeyLinkTemplate>
          <li class="bb-layout__horizontal-nav-item">
             <!-- ...etc -->
          </li>
        </ng-template>

        @if(journey.permissions) {
          <ng-container *bbIfEntitlements="journey.permissions">
            <ng-container *ngTemplateOutlet="journeyLinkTemplate" />
          </ng-container>
        } @else {
          <ng-container *ngTemplateOutlet="journeyLinkTemplate" />
        }
```

Maybe even better would be to speak with the web-foundation team to ask if they can change the `*bbIfEntitlements` directive so that if given a certain value, say `'*'`, then it automatically shows the content without checking any entitlements.  Then we could just do this:

```html
      <li
        *bbIfEntitlements="journey.permissions ?? '*'"
        class="bb-layout__horizontal-nav-item"
      >
        <!-- ... etc -->
      </li>
```

tcorral:
Good morning, @daiscog , @unlimitedjest .

What is the purpose and the status of this PR?

tcorral:
I close this PR as there are no updates since October 2024 and the author and the reviewer involved didn't reply to my message about the purpose of this PR, if it's needed afterwards you can always re-open it and solve the conflicts.

PR #372, opened 2025-01-20
WF-2024 Update text to translate to the new standards

tcorral:


package.json
```diff
@@ -61,6 +61,7 @@
     "angular-oauth2-oidc": "^17.0.0",
     "bignumber.js": "^9.1.1",
     "bootstrap": "^5.3.0",
+    "lodash": "^4.17.21",
```
ahnpnl, line 64: I think we don't need this one

ahnpnl (approved):

PR #378, opened 2025-01-21
WF-2018 Add examples of entitlements grouping - no foundation-ang update yet

tcorral:


package.json
```diff
@@ -61,6 +61,7 @@
     "angular-oauth2-oidc": "^17.0.0",
     "bignumber.js": "^9.1.1",
     "bootstrap": "^5.3.0",
+    "lodash": "^4.17.21",
```
RAJESH-GN, line 64: i think we can remove this as it was just to test pipeline issue

florianpasteur (approved):

PR #406, opened 2025-02-07
Use playwright mocks in Transactions e2e, update e2e tags

telmanagababov:
Use Playwright mocks in transactions e2e,
Add `useMocks` parameter to bypass mocks data in modelbank/ephemeral tests.

Update e2e tags to follow the Guideline https://backbase.atlassian.net/wiki/spaces/GUIL/pages/5259559552/E2E+tests+migration+guidelines


playwright.localhost.config.ts
```diff
@@ -58,12 +58,5 @@ export default defineConfig<TestOptions>({
...
   ],
-  webServer: [
-    {
-      command: 'npm run mock-server',
-      url: 'http://localhost:9999/dev-interface',
-      timeout: 30 * 1000,
-      reuseExistingServer: true,
-    },
-  ],
+  webServer: [],
```
vatsala-glory, line 61: do we still need this?

libs/transactions-journey/e2e-tests/specs/transaction-details.spec.ts
```diff
@@ -8,9 +8,14 @@ export function testTransactionDetails(
     'Transactions details',
     { tag: ['@e2e', '@transactions', '@transactions-details'] },
     () => {
-      test.beforeEach(async ({ detailsPage, detailsData }) => {
-        await detailsPage.navigate(detailsData.id);
-      });
+      test.beforeEach(
+        async ({ page, detailsPage, detailsData, detailsMocks }) => {
```
TomaszKasowicz, line 12: This is great.
The only thing i would consider is to move this whole code (which sets `page.route`) into fixture.
This way engineers will have full control when they want to override something on the app level

playwright.localhost.config.ts
telmanagababov: I just realised the we need this line to override the `baseConfig` web server configuration, I reverted it back  😀

libs/transactions-journey/e2e-tests/fixture/mocks.ts
```diff
@@ -0,0 +1,60 @@
...
+import { readFileSync } from 'node:fs';
+
+const accessControlMocks = {
+  success: {
+    '**/accessgroups/users/permissions/summary': JSON.parse(
+      readFileSync(
+        './mock-server/mocks/golden-sample-app/accessgroups/users_permission_summary.json',
+        'utf-8'
+      )
+    ).responses.ok.data,
```
daiscog, line 11: I'm not sure it's good to have a tight coupling here with the format of ng-apimock config files.  Also, feels a little heavy-handed to read all these json files into memory (could be hundreds in a large app) and store them statically in one big object.

I'd rather we had the tests define their own data, rather than rely on the config of an unrelated mock server.  Engineers should be free to edit the mock server config/data without compromising the tests.

Also, we need to make sure that nx is aware of any external dependencies this test project has, so that tests are properly rerun via `nx affected` if the files they depend on change.  Having mock data come from files within the same project will make that easiest.

There are options though if we want to explore sharing mock data between tests and the mock server.  E.g., we can use `route.fulfill({ path: 'path/to/response.json' })` in the tests and use `"responses": { "ok": { "file": "path/to/response.json" ...` in the ng-apimock config file.  But as noted if we do this we must make sure we configure the e2e target inputs so nx knows the tests are affected by these external files.

telmanagababov: Updated

TomaszKasowicz (approved):

PR #421, opened 2025-03-03
Add Aider action & workflows to generate a PR from an Issue

tcorral:


.github/workflows/aider-issue-to-pr.yml
```diff
@@ -0,0 +1,225 @@
...
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.ref }}
+
+      - name: Create a new branch
+        uses: actions/github-script@v7
+        id: create_branch
+        with:
+          script: |
```
RAJESH-GN, line 63: can we use this tomas just to keep the code more readable in yml file 
```
 - name: Read script from file
        id: read_script
        run: echo "::set-output name=script_content::$(cat .github/scripts/create_branch.js)"

      - name: Create a new branch
        uses: actions/github-script@v7
        id: create_branch
        with:
          script: ${{ steps.read_script.outputs.script_content }}
```

RAJESH-GN: for all the scripts in the files. just makes it a bit cleaner and readable and not crowded

tcorral: It's just a POC that I want to test, we can put more effort and improve it if we decide to use it in the future.

RAJESH-GN: Sure tomas but for me it is getting into main so i would ideally wish to have anything going into main to be clean. but okie.

PR #466, opened 2025-04-22
Initial migration with tests fixed

tcorral:


tcorral:
After composable apps have been merged we need to start over to resolve all the issues and get the improvements before starting the migration.

PR #469, opened 2025-04-24
Update pull-request.yanl for tests

RomanIovlev:
1. Split tests to categories: Mocks, Responsiveness, Accessibility, Sandbox
2. Make tests run in parallel
3. Add constant for retention days
4. Get Playwright version from package.json and pass it to Playwright Docker Image
5. Add Upload Allure and Accessibility report
6. Add cache for npm packages and playwright browsers


.github/workflows/pull-request.yml
```diff
@@ -52,18 +83,57 @@ jobs:
         with:
           npm-auth-token: ${{ secrets.NPM_TOKEN }}
 
+      - name: Cache Playwright browsers
+        if: ${{ success() || failure() }}
```
tcorral, line 87: Why do we check for success or failure and we execute it in both cases?

.github/workflows/pull-request.yml
```diff
@@ -8,10 +8,30 @@ on:
     branches:
       - main
 
+env:
+  RETENTION_DAYS: 7
+
+permissions:
+  contents: read
```
tcorral, line 15: Why do you need to give this permissions?

.github/workflows/pull-request.yml
```diff
@@ -8,10 +8,30 @@ on:
     branches:
       - main
 
+env:
+  RETENTION_DAYS: 7
```
tcorral, line 12: Is this used by the cache?

.github/workflows/pull-request.yml
```diff
@@ -52,18 +83,57 @@ jobs:
...
+          key: ${{ runner.os }}-playwright-${{ needs.get-playwright-version.outputs.version }}
+
       - name: Install browsers
         run: npx playwright install --with-deps chromium firefox webkit
         shell: bash
 
-      - name: Run Playwright e2e tests
-        run: HOME=/root npm run e2e
+      - name: Run Playwright e2e tests with mocks
+        if: ${{ success() || failure() }}
```
tcorral, line 98: Same question here.

.github/workflows/pull-request.yml
RomanIovlev: This is used to set how long to store testing reports: playwright report, allure report, accessibility report

.github/workflows/pull-request.yml
RomanIovlev: Maybe you right we can remove it for cache. This is required only for Test Steps so they will be marked as Failed but continue to next Test without failing whole job immediately

.github/workflows/pull-request.yml
RomanIovlev: This is required only for Test Steps so they will be marked as Failed but continue to next Test without failing whole job immediately
This is example how this look like: https://github.com/Backbase/golden-sample-app/actions/runs/14635571531/job/41065861168
Without that condition first test fail and will not run next step with tests
If we just add `continue-in-error: true` it will run ext test but step will be marked passed

.github/workflows/pull-request.yml
RomanIovlev: Unlike Artifacts cache is limited to 7 days by default but artifacts we should limit with retention-days (at least as I know)

.github/workflows/pull-request.yml
RomanIovlev: This is required because of get-playwright-version job
Without this Github shows security warning (please check in the history of Conversation)

.github/workflows/pull-request.yml
RomanIovlev: Yes because they are different types and not related Like responsiveness and accessibility

RomanIovlev: But they are all E2E

tcorral (approved):

TomaszKasowicz (approved):

PR #570, opened 2025-08-18
ci: add properties to skip e2e test folder for sonar analysis

vipin02:
THESE PROPERTIES WERE ADDED AS A REFERENCE FOR PROJECTS, AS THESE FILES DO NOT DIRECTLY AFFECT THE ACTUAL BUILD


tcorral:
Not relevant to Golden Sample App, we do not have any project in SonarCloud for this project and we do not have access to the one from Backbase either.

PR #582, opened 2025-09-03
Refresh: Part 1

daiscog:
First steps in refreshing the GSA.

- Revert to nx 20 and Angular 19 for now so we can remove use of overrides to circumvent peer dep issues
- eslint 9 and new "flat config"
- Switch to Angular's default esbuild-based application builder
- Fix issues preventing the app from starting
- Better standalone bootstrapping and environment file structure
- Fix broken dev server proxy configs
- Move all playwright config into the e2e test project (remove global-level config files) and revamp how different e2e testing scenarios are configured and executed

Lots more changes still to come in future PRs, but don't want to keep these changes out of main for too long to avoid risking awkward merge conflicts

Future work:

- Adopt BB common workflows
- More improvements to e2e tests including readme
- Remove use of nested nx projects for internal entry points (and bring in sheriff for module dependency linting)
- Standalone journeys with new journeyFactory API
- Clean up of any other obsolete code...


.cursor/mcp.json
```diff
@@ -0,0 +1,8 @@
+{
```
cheshirsky, line 1: Why do we need this one?

daiscog: It was auto-generated by the nx console plugin in cursor.  We have two options:

- gitignore this and let individual developers manage their own MCP configs
- commit and have a single shared MCP config for all team memebers

I think it's unlikely people in the same team working on the same project will want different MCP configurations for cursor, and sharing it improves governance, so I've sided with the latter.

TomaszKasowicz (approved):

PR #596, opened 2025-11-16
feat: upgrade to angular 20 and migrate to standalone architecture

tcorral:


apps/golden-sample-app/src/app/error-page/error-page.component.ts
```diff
@@ -1,11 +1,13 @@
...
 import { Component, inject } from '@angular/core';
+import { CommonModule } from '@angular/common';
 import { Router } from '@angular/router';
+import { ErrorCommonStateModule } from '@backbase/ui-ang/common-error-state';
 
 @Component({
   selector: 'app-error-page',
   templateUrl: './error-page.component.html',
-  standalone: false,
+  imports: [CommonModule, ErrorCommonStateModule],
```
daiscog: I don't think we need to import Angular's `CommonModule` in this component - I don't see any uses of anything from it in the [template](https://github.com/Backbase/golden-sample-app/blob/e8c0246e015025225666dbfa0f9248593a2f7fdf/apps/golden-sample-app/src/app/error-page/error-page.component.html)

libs/ach-positive-pay-journey/src/lib/ach-positive-pay-journey-shell.config.ts
```diff
@@ -0,0 +1,20 @@
+import { achPositivePayDefaultRoutes } from './ach-positive-pay-journey.routes';
+import { Routes, provideRoutes } from '@angular/router';
+
+/**
+ * Provides the default routes for ACH Positive Pay journey.
+ * For standalone applications, use provideAchPositivePayRoutes() in app config.
+ */
+export function provideAchPositivePayRoutes(
```
daiscog: Where is this used?  Doesn't look like the usual pattern for providing journey routes.

Let's delete it if it's unused.

libs/journey-bundles/transactions/src/lib/transactions.bundle.ts
```diff
@@ -44,38 +45,32 @@ const routes: Routes = transactionsJourney(
...
-      provide: TransactionsJourneyConfiguration,
-      useFactory: (configService: TransactionsConfigService) => {
-        const config = configService.getJourneyConfig();
-        const journeyConfig = new TransactionsJourneyConfiguration();
-        journeyConfig.pageSize = config.pageSize;
-        journeyConfig.slimMode = config.slimMode;
-        return journeyConfig;
-      },
-      deps: [TransactionsConfigService],
+export const TRANSACTIONS_PROVIDERS = [
```
daiscog: I'm not sure any of these are needed?  Looks like they're all provided in the journey function above in the new standalone way?  E.g., `withConfig` replaces the need for explicitly providing `TransactionsJourneyConfiguration`, etc.

`TransactionsConfigService` doesn't look like it's used at all, except in the factory functions below.

`TRANSACTIONS_JOURNEY_CONFIG` and `TransactionsJourneyConfiguration` seem to be duplicate providers for the same thing - looks like TRANSACTIONS_JOURNEY_CONFIG is never actually injected anywhere.

libs/shared/feature/user-context/src/lib/shared-user-context.providers.ts
```diff
@@ -0,0 +1,7 @@
+import { SharedUserContextService } from './shared-user-context.service';
+import { SharedUserContextGuard } from './shared-user-context.guard';
+
+export const SHARED_USER_CONTEXT_PROVIDERS = [
+  SharedUserContextService,
+  SharedUserContextGuard,
+];
```
daiscog: As per earlier comments, I'd delete this as these are both `providedIn: 'root'`

libs/transactions-journey/src/lib/transactions-journey-shell.config.ts
```diff
@@ -0,0 +1,70 @@
...
+export interface TransactionsJourneyConfig {
+  routes?: Routes;
+  extensionSlots?: TransactionsJourneyExtensionsConfig;
+}
+
+/**
+ * Provides the default configuration for Transactions Journey.
+ * For standalone applications, use provideTransactionsJourney() in app config.
+ */
+export function provideTransactionsJourney({
```
daiscog: Doesn't look like this is used anywhere

libs/transfer-journey/src/lib/transfer-journey-shell.config.ts
```diff
@@ -0,0 +1,85 @@
...
+  MakeTransferRouteTitleResolverService,
+} from '@backbase/transfer-journey/internal/data-access';
+import { TRANSLATIONS } from '@backbase/transfer-journey/internal/shared-data';
+
+export const defaultTransferRoute: Route = {
+  path: '',
+  loadComponent: () =>
+    import('./transfer-journey.component').then(
+      (m) => m.TransferJourneyComponent
+    ),
```
daiscog: Again, given this whole routes array itself will be lazy-loaded, I'm not sure it's worth lazy-loading the individual components within it too.

libs/transfer-journey/src/lib/transfer-journey-shell.config.ts
```diff
@@ -0,0 +1,85 @@
...
+      canActivate: [MakeTransferJourneyStoreGuard],
+    },
+  ],
+};
+
+/**
+ * Provides the default configuration for Transfer Journey.
+ * For standalone applications, use provideTransferJourney() in app config.
+ */
+export function provideTransferJourney(
```
daiscog: I don't see where this is being used either

libs/journey-bundles/transfer/src/lib/route.ts
```diff
@@ -1,11 +1,89 @@
...
-      (m) => m.TransferJourneyBundleModule
-    ),
+  loadChildren: async () => {
+    const [journeyModule, featureModule, dataAccessModule] = await Promise.all([
+      import('@backbase/transfer-journey'),
+      import('@backbase/transfer-journey/internal/feature'),
+      import('@backbase/transfer-journey/internal/data-access'),
+    ]);
+
+    return [
```
daiscog: This also doesn't look like the usual patterns for loading a journey - again we'd normally have a `transfer-journey-bundle` ts file that we lazy-load alone for loadChildren, and that file will export the routes as a default export, with the routes coming from the journey's entry function in the `@backbase/transfer-journey` dir.  We should never be importing directly from `internal` entry points of other packages.

RAJESH-GN (dismissed):

libs/journey-bundles/ach-positive-pay/src/lib/ach-positive-pay-journey-bundle.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * The routes include all necessary components and guards from the journey library.
+ */
+const routes: Routes = achPositivePayDefaultRoutes;
+
+// Default export for Angular lazy loading
+export default routes;
+
+// Named export for backward compatibility and testing
+export const ACH_POSITIVE_PAY_ROUTES = routes;
```
daiscog: There is absolutely no need to export the same thing as a default export and as a named export.  Default alone is fine.

libs/journey-bundles/transactions/src/lib/transactions.bundle.ts
```diff
@@ -3,79 +3,60 @@ import {
...
-  }
-}
+// Create wrapper route that includes the required providers
+// This ensures providers are available when the journey is lazy-loaded
+const routes: Routes = [
+  {
+    path: '',
+    providers: journeyProviders,
+    children: journeyRoutes,
+  },
```
daiscog: We can use `withProviders` from foundation-ang inside the `transactionsJourney(...)` call above to avoid needing this wrapper route (this is what `withProviders` is designed for)

libs/journey-bundles/transactions/src/lib/transactions.bundle.ts
```diff
@@ -3,79 +3,60 @@ import {
...
+const journeyProviders = [
+  {
+    provide: TransactionsJourneyConfiguration,
+    useValue: journeyConfig,
+  },
+  TransactionsRouteTitleResolverService,
+  {
+    provide: TRANSACTION_EXTENSIONS_CONFIG,
+    useValue: extensionsConfig,
+  },
```
daiscog: This provider for `TRANSACTION_EXTENSIONS_CONFIG` isn't needed as that's what `withExtensions` does below

libs/journey-bundles/transfer/src/lib/transfer-journey.bundle.ts
```diff
@@ -0,0 +1,60 @@
...
+  maskIndicator: false,
+  maxTransactionAmount: 100,
+  slimMode: false,
+};
+
+// Providers needed by the journey's internal services
+const journeyProviders = [
+  provideJourneyTracker({
+    journeyName: 'transfer',
+  }),
```
daiscog: If something is needed internally in the journey, that should normally be provided in the journeyFactory call.  E.g.:

```ts
export const {
  makeTransferJourney,
  //...
} = journeyFactory({
  journeyName: 'makeTransferJourney',
  //...
  providers: [
    provideJourneyTracker({
      journeyName: 'transfer',
    }),
    MakeTransferAccountHttpService, // even better, this one could be `providedIn: 'root'`
  ]
});
```

libs/transactions-journey/src/lib/transactions-journey-shell.config.ts
```diff
@@ -0,0 +1,11 @@
+import { InjectionToken } from '@angular/core';
+
+export type TransactionsJourneyExtensionsConfig = Record<string, unknown>;
+export const TRANSACTION_EXTENSIONS_CONFIG =
```
daiscog: Looks like this is a duplicate of the existing token [here](https://github.com/Backbase/golden-sample-app/blob/a7653afcb6d0fe4958f31a58915543a474d053fa/libs/transactions-journey/internal/feature-transaction-view/src/lib/extensions/config.ts#L9) - this is probably why you're having to provide it twice in the journey bundle, once for each token.

libs/transfer-journey/src/lib/transfer-journey.ts
```diff
@@ -0,0 +1,123 @@
...
+  maxTransactionAmount: 100,
+  slimMode: true,
+};
+
+// Configuration Token
+export const MAKE_TRANSFER_JOURNEY_CONFIG =
+  new InjectionToken<MakeTransferJourneyConfig>('MAKE_TRANSFER_JOURNEY_CONFIG');
+
+// Communication Service Token
+export const MAKE_TRANSFER_JOURNEY_COMMUNICATION_SERVICE =
```
daiscog: I also don't see this being injected anywhere either - but I do see we're injecting the abstract `MakeTransferCommunicationService` class.

As this is a service, not a plain config value, I think it's better in this instance to delete this token and use `communicationService: MakeTransferCommunicationService` in the `journeyFactory` call below.

daiscog (approved):

PR #599, opened 2025-12-09
[DO NOT MERGE] Feature/bbnc 64 web ai sdlc test

cheshirsky:
PR represents engineer following  the [playbook](https://github.com/Backbase/golden-sample-app/blob/feature/BBNC-64-ai-sdlc-web/docs/playbooks/JIRA-001-playbook.md) on generating code. The playbook describes the following algorythm:
image (2)

Relevant files are:
1. All source code (tests + actual code)
2. `docs/specs` folder (which is "part 1: specs" intermediate artefacts)
3. `docs/validation` folder (represents is "part 3" intermediate artefacts)


libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.html
```diff
@@ -51,20 +72,46 @@
...
+              Transactions empty state subtitle - 'There are no transactions for
+              this account.'|This string is used as the subtitle for the empty
+              state when no transactions are available.
+              @@transactions.empty-state.subtitle"
+          >
+          </bb-empty-state-ui>
+        </div>
+
+        <!-- Transactions List -->
+        <ul *ngIf="items.length > 0" class="bb-list">
```
JonnyHeavey: We should prefer control flow in our templates as structural directives like ngIf and ngFor are deprecated. This would be particularly effective for the loop in this template, where you can use [at]empty as well.

Though appreciate this code change is inkeeping with existing template style, so this could be deferred if you choose.

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.html
```diff
@@ -51,20 +72,46 @@
...
+          <bb-empty-state-ui
+            title="No transactions"
+            i18n-title="
+              Transactions empty state title - 'No transactions'|This string is
+              used as the title for the empty state when no transactions are
+              available for the selected account.
+              @@transactions.empty-state.title"
+            subtitle="There are no transactions for this account."
+            i18n-subtitle="
+              Transactions empty state subtitle - 'There are no transactions for
```
JonnyHeavey: Its unconventional to include the actual default sentence in the i18n key description

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.ts
```diff
@@ -71,12 +101,47 @@ export class TransactionsViewComponent {
...
+   */
+  ngOnInit(): void {
+    this.initDefaultAccountSelection();
+  }
+
+  /**
+   * Sets up default account selection logic.
+   * RULE: Navigates to first account if no account or invalid account in URL.
+   */
+  private initDefaultAccountSelection(): void {
```
JonnyHeavey: private methods should be located after public / protected ones

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.module.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+      );
+
+      // Act
+      const hasAccountSelector = moduleImports.some(
+        (importedModule) =>
+          importedModule === AccountSelectorModule ||
+          (importedModule as { name?: string })?.name === 'AccountSelectorModule'
+      );
+
+      // Assert
```
JonnyHeavey: These inline comments on test case structure are redundant

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.module.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+
+      // Act
+      const module = TestBed.inject(TransactionsViewModule);
+
+      // Assert
+      expect(module).toBeDefined();
+    });
+
+    // Edge case: Module exports are available after import
+    it('should_export_components_when_module_is_imported', () => {
```
JonnyHeavey: Poor formatting on test descriptions. There is no need for these to be underscored.

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.ts
```diff
@@ -71,12 +101,47 @@ export class TransactionsViewComponent {
...
+    })
+      .pipe(
+        // RULE: Only process once when accounts are loaded
+        filter(({ accounts }) => accounts.length > 0),
+        take(1),
+        takeUntilDestroyed(this.destroyRef)
+      )
+      .subscribe(({ selectedAccount, accounts }) => {
+        // RULE: Navigate to first account if no valid selection
+        if (!selectedAccount) {
```
JonnyHeavey: Having conditions inside a subscription handler is often a smell. Consider using filter operator to apply this logic instead (filter(({ selectedAccount }) => !selectedAccount))

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.html
```diff
@@ -11,6 +11,27 @@
     </div>
 
     <div class="bb-card__body">
+      <!-- Account Selector -->
```
JonnyHeavey: This is subjective, but some of the inline commenting in this template is unhelpful in my view. Here we have a comment which is repeating what the well-named component selector is already communicating.

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.ts
```diff
@@ -31,6 +32,35 @@ export class TransactionsViewComponent {
     map((params) => params.get('account'))
   );
 
+  /**
```
JonnyHeavey: This tsdoc describes the implementation (which itself is clear) rather than what the property itself is

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.html
cheshirsky: +1. Right now old directives were used as LLM is forced to "follow repo standards", but we can influence it via ADRs

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.html
cheshirsky: Agreed, I'll check if it can be addressed via adding it to the ADR

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.html
daiscog: I agree.  Most code comments are prone to becoming stale over time, so become a distraction.  If code is well-written, explanatory comments aren't needed.

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.html
```diff
@@ -11,6 +11,27 @@
...
+      <div class="bb-block bb-block--md">
+        <bb-account-selector-ui
+          data-role="transactions-view__account-selector"
```
daiscog, line 27: We recommend using accessibility (role-based) selectors in our tests over `data-role` attributes, with `data-testid` only as a fallback when a role-based selector is too cumbersome.

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.html
```diff
@@ -51,20 +72,46 @@
         </bb-badge-ui>
       </div>
 
-      <ul *ngIf="transactions$ | async as items; else loading" class="bb-list">
-        <li
-          *ngFor="let item of items | filterTransactions : filter"
-          class="bb-list__item"
-          data-role="transactions-view__item-container"
+      <ng-container *ngIf="transactions$ | async as items; else loading">
```
daiscog: `*ngIf` is deprecated.  We should be using `@if`.

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.html
```diff
@@ -51,20 +72,46 @@
...
+        </div>
+
+        <!-- Transactions List -->
+        <ul *ngIf="items.length > 0" class="bb-list">
+          <li
+            *ngFor="let item of items | filterTransactions : filter"
+            class="bb-list__item"
+            data-role="transactions-view__item-container"
+          >
+            <a class="list-link" [routerLink]="['./', item.id]">
```
daiscog: Nit-pick, but I'm pretty sure routerLink here can be simplified to `[routerLink]="item.id"` since the routerLink directive is relative to the current activated route by default

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.ts
```diff
@@ -71,12 +101,47 @@ export class TransactionsViewComponent {
     private readonly router: Router,
     private readonly transactionsService: TransactionsHttpService,
     private readonly arrangementsService: ArrangementsService,
+    private readonly destroyRef: DestroyRef,
```
daiscog: We should prefer `inject` over constructor injection

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.ts
```diff
@@ -71,12 +101,47 @@ export class TransactionsViewComponent {
...
+  ngOnInit(): void {
+    this.initDefaultAccountSelection();
+  }
+
+  /**
+   * Sets up default account selection logic.
+   * RULE: Navigates to first account if no account or invalid account in URL.
+   */
+  private initDefaultAccountSelection(): void {
+    combineLatest({
```
daiscog: This feels like a very convoluted way of selecting the first account by default, and also adds a navigation to the history stack (I'd be suspicious about what effect using the browser back button would have).

A much simpler approach would be for the `selectedAccount$` observable to emit the first account if the `accountId$` observable emits undefined (or even any invalid value) then we won't need to have this programmatic navigation to add the first account id to the query params

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.spec.ts
```diff
@@ -195,4 +298,652 @@ describe('TransactionsViewComponent', () => {
...
+      // Arrange
+      setup(snapshot);
+      arrangements$$.next(mockArrangements);
+
+      // Act
+      const accounts = await firstValueFrom(getAccounts$());
+
+      // Assert
+      expect(accounts[0].number).toBe('123456789');
+    });
```
daiscog: These tests could be collapsed into one test that does 

```ts
expect(accounts[0]).toStrictEqual({
  // ... all of the properties here
})
```

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.spec.ts
```diff
@@ -195,4 +298,652 @@ describe('TransactionsViewComponent', () => {
...
+
+    /**
+     * Helper to access accounts$ observable.
+     * Uses type assertion since property is being added in S2.
+     */
+    const getAccounts$ = () =>
+      (fixture.componentInstance as unknown as { accounts$: Observable<AccountSelectorItem[]> }).accounts$;
+
+    // Happy path: accounts$ emits mapped arrangement data
+    it('should_emit_mapped_accounts_when_arrangements_are_loaded', async () => {
```
daiscog: I'm not a fan of the underscores instead of spaces in the test names - is there a reason for these?

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.spec.ts
```diff
@@ -195,4 +298,652 @@ describe('TransactionsViewComponent', () => {
...
+        title: 'Transactions',
+      },
+    };
+
+    /**
+     * Helper to access accounts$ observable.
+     * Uses type assertion since property is being added in S2.
+     */
+    const getAccounts$ = () =>
+      (fixture.componentInstance as unknown as { accounts$: Observable<AccountSelectorItem[]> }).accounts$;
```
daiscog: Are these "helpers" still needed? They could mask genuine typing errors, and the JSDoc description sounds stale.

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.spec.ts
```diff
@@ -195,4 +298,652 @@ describe('TransactionsViewComponent', () => {
...
+
+      // Act
+      const accounts = await firstValueFrom(getAccounts$());
+
+      // Assert
+      expect(accounts).toHaveLength(0);
+    });
+  });
+
+  describe('selectedAccount$ observable (S2)', () => {
```
daiscog: Just a thought: For non-trivial observables, I like to use marbles testing so we can assert how an observable will change over time, in reaction to new values being emitted by the source observables.

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.module.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+}
+
+/**
+ * Type for module class with Ivy metadata.
+ */
+interface ModuleWithIvyMetadata {
+  ɵmod?: IvyModuleDef;
+  ɵinj?: IvyModuleDef;
+}
+
```
daiscog: Why would we care about Angular's internal APIs?

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.module.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+      // Arrange
+      const moduleSetup = () => {
+        TestBed.configureTestingModule({
+          imports: [TransactionsViewModule],
+        });
+      };
+
+      // Act & Assert
+      expect(moduleSetup).not.toThrow();
+    });
```
daiscog: This feels like a really low-value test.  If using the module throws then all other tests that use it will fail.

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.module.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+      const moduleImports = getModuleImportsIvy(
+        TransactionsViewModule as unknown as ModuleWithIvyMetadata
+      );
+
+      // Act
+      const hasAccountSelector = moduleImports.some(
+        (importedModule) =>
+          importedModule === AccountSelectorModule ||
+          (importedModule as { name?: string })?.name === 'AccountSelectorModule'
+      );
```
daiscog: I'm not sure I see the point of this test either.  We should be testing business logic and behaviours not Angular boilerplate code.

We never need to assert that one module imports another  - the Angular compilation will fail if we have missing module imports that our components need.

(Same for all the other similar tests)

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.module.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+      TestBed.configureTestingModule({
+        imports: [TransactionsViewModule],
+      });
+
+      // Act
+      const module = TestBed.inject(TransactionsViewModule);
+
+      // Assert
+      expect(module).toBeDefined();
+    });
```
daiscog: Again, feels like we're testing Angular constructs here, not our own code.  Tbh, the vast majority of `.module.ts` files do not need any unit tests at all, as they are 100% Angular boilerplate.  If it had things like `.forRoot` or other functions in the class that contained our own logic, or if it set up complicated providers, then it's worth adding tests for those.  But this entire file feels like a box-ticking exercise.

JonnyHeavey:
Overall, this is a reasonable start for an automated contribution, with the core of the implementation providing the required functionality. However, whilst the majority of the issues raised here are of a stylistic nature, they should be taken seriously to avoid degradation of codebase maintainability.

I'm sure with additional context (and perhaps automated reivew/verification loops) we can steer the agent to more preferable codestyle and patterns.

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.module.spec.ts
cheshirsky: Issue caused by instructions to use reasononing as inline comments to pivot code generation, prompt needs hardening. Added to my todo

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.ts
cheshirsky: Added to the todo

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.module.spec.ts
cheshirsky: Yeh, I am missing ut agent entirely on this prototype, as my point was to prove that tests improve code.

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.ts
cheshirsky: Added ADR to the list of missing artefacts

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.html
cheshirsky: yeh, leaked LCPs, sorry :)

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.html
cheshirsky: added to the todo

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.html
cheshirsky: added

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.html
cheshirsky: not sure if I need to reflect it anywhere, as it comes from the repo itself. Probably ADRs always prioritiesd over repo-level stuff.

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.spec.ts
cheshirsky: missing UT agent. Added

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.spec.ts
cheshirsky: agreed, naming should be part of UT agent clearly

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.spec.ts
cheshirsky: Is it a smell in your opinion, or style preference?

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.component.spec.ts
cheshirsky: Added

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.module.spec.ts
cheshirsky: mocking for the module test

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.module.spec.ts
cheshirsky: Good point

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.module.spec.ts
cheshirsky: fair, added to agent

libs/transactions-journey/internal/feature-transaction-view/src/lib/components/transactions-view/transactions-view.module.spec.ts
cheshirsky: added

cheshirsky:
@daiscog @JonnyHeavey added the list of outstanding issues with resolutions for them, WDYT?

PR #625, opened 2026-06-16
feat: added integration test example

srinivasjaithwala:


apps/golden-sample-app-e2e/src/data/mocks-data/index.ts
```diff
@@ -1 +1,2 @@
 export * from './transactions.mocks.api.data';
+export * from './transactions.integration.mocks.data';
```
edwardroth: Any reason we can't just use `./transactions.mocks.api.data` ?

srinivasjaithwala: updated accordingly.

edwardroth (approved):

